PR #20, opened 2014-10-08
Add docker support

carloslima:
Here it goes, @calumhalcrow :-)


calumhalcrow:
@carloslima I'm going to close this PR since we don't plan to have Docker-based environments. I'll leave the branch though.

PR #80, opened 2014-12-02
Added support for IM mapping for market research endpoint and importer. ...

kjubichan:
...The requests to IM mocked by VCR


calumhalcrow:
@jamesisaacs I feel this can be merged as-is, even though we haven't completed all IM work. The only thing that will change here in the future is the data we receive from the IM API. We won't need to make further code changes. Also, all changes only affect v2; v1 is unchanged. So merging this won't break anything that's relied on by our users.

PR #88, opened 2015-01-12
Tim/ccg squashed

calumhalcrow:


calumhalcrow:
@tmhammer due to the large amount of data files in these changes I have re-worked the commits so that the data is in one commit, and the code changes in another. I will add feedback to the code changes commit. Question: do you have further changes to push? If not, I will reset the tim/ccg branch to this so that you are working against the squashed commits.

tmhammer:
@calumhalcrow We shouldn't need any more changes for now.  Id has been removed from mapping, but I'm not sure how to shorten that entries block since we're concatenating arrays instead of inserting.

calumhalcrow:
@tmhammer regarding the concat: yeah, I see what you mean. Feel free to leave it as-is, or if you want to get fancy take a look at http://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-reduce

I'm going to close this PR now and open one against tim/ccg instead. I've reset that branch to this one (so that it has the squashed commits).

PR #89, opened 2015-01-14
Tim/ccg

calumhalcrow:


tmhammer:
@calumhalcrow  I think we'll eventually be writing those output files to an S3 bucket, so the writing to a local folder is just a stopgap for this functionality.  Having them appear in the repo made it easy for me to see how they looked.  But that being said, I guess we don't need to have this functionality on staging, since the files aren't being used by anything and I've just been running the import locally, and zipping and sending them to AJ (since there's no S3 yet).

calumhalcrow:
@tmhammer I'm not sure what happened, but I found that this PR had both the original commits you had created, the squashed commits that I had turned them into, a new commit from yourself, and finally a merge commit that seemed to have brought them all together. It should only contain the squashed commits and your new work.

Can you do the following on your local copy. It will sort you out so that you don't accidentally bring back those dead commits:
```
git fetch
git checkout tim/ccg
git reset --hard origin/tim/ccg
```

calumhalcrow:
@tmhammer I'm still seeing deleted and added files when I do a local import. Seems that the date in the filename changes:
```
~/GovWizely/webservices [tim/ccg|✚121…259⚑7]
17:49 $ git status
On branch tim/ccg
Your branch is up-to-date with 'origin/tim/ccg'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    data/country_commercial_guides/china/2015-01-16-agricultural-best-prospects.md
        deleted:    data/country_commercial_guides/china/2015-01-16-agricultural-overview.md
...
        deleted:    data/country_commercial_guides/colombia/2015-01-16-defense-overview.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        data/country_commercial_guides/china/2015-01-17-agricultural-best-prospects.md
        data/country_commercial_guides/china/2015-01-17-agricultural-overview.md
```

calumhalcrow:
@tmhammer I've been looking at this some more, and I can see how to adjust the importer in such a way that no untracked data files need be generated. I haven't got to actually making the necessary code change though. I'll get to this on Tuesday. If it's not ok to wait until I can do this let me know.

calumhalcrow:
@tmhammer I have now gone over your work in detail. Initially I intended to make some changes to the importer, but after my review I found myself confused by a few things, so I think it’s better if I ask you for clarification on them, and make suggestions about an alternative approach you should take rather than do all the work myself.

There’s a couple of things I want to discuss:

1: What appears to be a two-step import approach:

There are source docs for four countries: CN, CO, BR and AR. For CN/CO, there is one YAML file and one MD file per country. Combined, these two files contain all documents for the country being imported. For BR/AR, there is one MD file for each individual document that is to be imported. The code in your importer suggests that for CN/CO, you are trying to convert the “all-in-one” files to “1-doc-per-file” files, similar to the BR/AR format.

I assumed that you would do this “step 1: all-in-one to 1-doc-per-file” first, then do the actual import against the 1-doc-per-file files. In reality, that doesn’t seem to be the case. If I comment out the lines that write the 1-doc-per-file files to disk, the importer still works as expected. This leads to my first point of confusion: I’m not sure why you’re taking the trouble to write these files if they are not used during the import process.

Secondly, although several BR/AR 1-doc-per-file files are present, they are not touched by the importer. For example, after doing the import, I only see CN and CO docs in the index. If you are not importing them, why are they present?

My assumption from all of this is that you wish to do a two-step import. Step 1 will convert the large all-in-one files to 1-doc-per-file files. Step 2 will import the 1-doc-per-file files. Is this what you intend to do? If so, we should make these code changes before considering the PR good to merge.

2: Approach to importing MD

I noticed you are extracting content straight from the raw MD string, using string methods like seek and pos. You’ll generally have a better time if you use a gem to parse the markdown, then extract the information you need from the parsed information. Something like:
```
md_str = open("#{Rails.root}/data/country_commercial_guides/markdown/#{@md_file}”).read
md_doc = Kramdown::Document.new(md_str)
html_doc = Nokogiri::HTML(md_doc.to_html)
relevant_content = html_doc.css(‘#market-overview’).text
```
The only problem with this approach right now is that it seems the large MD files do not convert to HTML correctly. The screenshot highlights the problem:

screen shot 2015-01-20 at 9 52 39 am

My suggestion is to make sure the MD docs are formatted correctly, and use Kramdown/Nokogiri to extract the info you need.

--

I'd like to sort these two issues out before merging this PR.

tmhammer:
@calumhalcrow   Ok, I'll look in to converting the Markdown to HTML, that definitely seems like a better method.  The broken-down Markdown files are not doing anything in the import process, they're just output files for AJ to use on the front end for the report generator.  So we don't really need to have this functionality on staging... if we need new output files we could just uncomment a few lines and run locally, then add the big Markdown and YAML files to staging to get the JSON entries.

tmhammer:
@calumhalcrow  I updated the import method to parse using Nokogri.  I think the dates on the snippet file names are ok given they reflect when the files were created, but it's kind of arbitrary anyways since the only reason we need to do this is so Jekyll can render them properly.  So the branch is good to go if you want to make any changes in regards to handling the output files.

calumhalcrow:
@tmhammer I have finished switching over the importer to read from https://github.com/GovWizely/ccg rather than from source files stored in data/. Do you want to review the changes and merge if happy?

tmhammer:
Cool, yeah I'll try to get this reviewed today.

PR #86, opened 2015-01-11
Auto recreate/import changed indices after deployment

calumhalcrow:
[#85422738]

PR #90, opened 2015-01-18
CSL: exact matches on name/alt_names fields

carloslima:


app/models/screening_list/mappable.rb
```diff
@@ -36,7 +36,11 @@ def self.included(klass)
                                          name: { type: 'string', analyzer: 'standard_asciifolding_nostop' },
                                          sort: { type: 'string', analyzer: 'keyword_lowercase' } } },
 
-            alt_names:               { type: 'string', analyzer: 'standard_asciifolding_nostop' },
+            alt_names:               { type:   'multi_field',
+                                       fields: {
+                                         alt_names: { type: 'string', analyzer: 'standard_asciifolding_nostop' },
+                                         sort:      { type: 'string', analyzer: 'keyword_lowercase' } } },
```
carloslima: I went with the same naming as the 'name' field and used 'sort'. Is that the best name for this?

calumhalcrow: Before this, "sort" was only used for sorting, and the name made sense. Now, we're using it for two things: sorting and exact matching, so a different name might be better (like "keyword" or something).

BTW, while you're here can you switch this multi_field definition to use the [updated syntax](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_multi_fields.html)? I've changed it on [this forthcoming PR](https://github.com/GovWizely/webservices/pull/86), but you might as well change it here, if only to become more aware of the deprecation.

PR #92, opened 2015-01-25
Fix missing country lookups

carloslima:


lib/importer.rb
```diff
@@ -38,21 +38,39 @@ def lookup_country(country_str)
   end
 
   def normalize_country(country_str)
```
carloslima: Any suggestion on how to make this long (and possibly ever growing) method better?

calumhalcrow: I guess we could examine the innards of IsoCountryCodes and copy the techniques they use, if they seem useful. We could also possibly remove the regexs and put the mapping in a config file.

carloslima: I liked the config file idea. What do you think of that implementation?

PR #93, opened 2015-01-26
Import directly from S3.

tmhammer:
Revert "Import directly from S3."

This reverts commit e16f913a39ba6e12e38e464f779386618f164076.

Finished updates.

Delete unnecessary files.


tmhammer:
@calumhalcrow  Build is passing now so this should be ready for review.

calumhalcrow:
:+1:

app/importers/sharepoint_trade_article_data.rb
```diff
@@ -44,24 +43,38 @@ class SharepointTradeArticleData
     trade_url:             [],
   }.freeze
 
-  def initialize(resource = ENDPOINT)
-    @resource = resource
+  def initialize(s3 = nil)
+    @s3 = s3 || Aws::S3::Client.new(
+      access_key_id: ENV['AWS_ACCESS_KEY_ID_ARTICLES'], 
+      secret_access_key: ENV['AWS_SECRET_ACCESS_KEY_ARTICLES'])
```
calumhalcrow: @tmhammer can you send me what these environment variables should be set to? They'll be needed for staging/production.

spec/spec_helper.rb
```diff
@@ -18,6 +18,7 @@
   c.hook_into :webmock
   c.debug_logger = File.open('log/vcr.log', 'w')
   c.configure_rspec_metadata!
+  c.allow_http_connections_when_no_cassette = true
```
calumhalcrow: This isn't the way to deal with VCR errors. What's happening is VCR is telling you "hey, a spec is trying to access the network, which isn't ok. Please fix it." and this code change is answering "that's ok, just go ahead and access the network".

On top of that, the network activity which is happening is kinda interesting. The code is making requests to 169.254.169.254, which is a "dynamically configured link-local addresses", which is private to the local network and won't ever be accessible over the internet like other IP addresses are. This [server fault Q&A](http://serverfault.com/questions/427018/what-is-this-ip-address-169-254-169-254) explains the situation quite well. This specific IP address is used internally by AWS. All-in-all, I think something is not right here; maybe the use of the S3 gem's interface. I'd happily debug this, but I need the access creds as requested above.

To be clear though: we need to remove this from the VCR config. I don't want our specs taking to the network during their run.

PR #96, opened 2015-02-10
Tim/tariffs new sources

tmhammer:


jamesisaacs:
@tmhammer please change the code to use the 2-digit country codes are used as values for the *countries* parameter. I should've set it up this way from the start.

.gitignore
```diff
@@ -16,3 +16,5 @@
 .*.swp
 /tags
 /capistrano
+
+dump.rdb
```
calumhalcrow: What is this file?

tmhammer: I think the dump.rdb is a Redis data storage thing that gets created on the fly... so I must have left this here from when we were using Sidekiq.

calumhalcrow: In that case please remove this change :-)

calumhalcrow:
I wonder how long it takes to import all tariff rates sources. We may need sidekiq, even though we're not importing the "big file".

calumhalcrow:
@tmhammer I have just rebased this branch onto master, to pick up recent changes there. Before doing any further work on this branch (and assuming you don't have any local changes), please reset your local branch to the remote:
```
git fetch
git checkout tim/tariffs-new-sources
git reset --hard @{u}
```
[If you do have local changes, wrap them up in a commit (`git commit -amWIP`), copy the commit's sha to your clipboard, do the steps above, then cherry-pick the commit (`git cherry-pick [the sha]`).]

app/importers/tariff_rate/morocco_data.rb
```diff
@@ -0,0 +1,9 @@
+module TariffRate
+  class MoroccoData
+    include ::Importer
+    include TariffRate::Importer
+
+    self.default_endpoint = 'FTA_Morocco_Data.csv'
+    self.country_code = 'MA'
```
calumhalcrow: Since tariff rates sources are by country, is there any reason why we need this `country_code` attribute in addition to the model's `source[:code]` attribute?

config/sidekiq.yml
```diff
@@ -0,0 +1,2 @@
+---
+:concurrency: 10
```
calumhalcrow: Can you remove this file also, since we're not using sidekiq?

calumhalcrow:
@tmhammer in addition to my feedback, can you do a `rubocop -a` on these changes?

calumhalcrow:
FYI full import on staging took 1h48m.

PR #97, opened 2015-02-16
MRL specs adjusted for new v2 results format.

calumhalcrow:


calumhalcrow:
@jamesisaacs this is just an adjustment to the MRL specs, which were previously marked as pending, pending confirmation of the v2 results format. Since the format has been on master for a long time now, this change is long overdue.

PR #99, opened 2015-02-24
SBA trade events ssl connection error on production

calumhalcrow:
[#88863548]

### DIFF
--- a/app/importers/trade_event/sba_data.rb
+++ b/app/importers/trade_event/sba_data.rb
@@ -1,4 +1,4 @@
-require 'open-uri'
+require 'net/https'
 
 module TradeEvent
   class SbaData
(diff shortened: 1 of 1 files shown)


app/importers/trade_event/sba_data.rb
```diff
@@ -60,10 +60,21 @@ def import
...
 
+    # :nocov:
+    def open_with_tlsv1(url)
+      uri = URI(url)
+      Net::HTTP.start(uri.host, uri.port, use_ssl: true, ssl_version: :TLSv1) do |http|
+        request = Net::HTTP::Get.new(uri, @header)
+        http.request(request).body
+      end
+    end
+    # :nocov:
```
calumhalcrow: This worked on Ubuntu 14.04:
```
curl -g  --tlsv1.0 "https://www.sba.gov/event-list/views/new_events_listing?display_id=services_1&filters[event_topic][value]=6&limit=100&offset=0"
```

jamesisaacs:
I'm bypassing staging for this PR since it contains a bugfix that is only relevant to the version of OpenSSL used on production. The code has been tested on an environment that is similar to production.

PR #100, opened 2015-02-26
Import full XML

tmhammer:
Update import, remove sidekiq worker

Import MOD entries from both sources

Fix specs

Cleanup


tmhammer:
@calumhalcrow , this work is complete but I'm open to suggestions on how to best handle the separate imports.  Right now you can run the full import by entering "bundle exec rake ita:import[TradeLead::Fbopen]" and the partial import by entering "bundle exec rake ita:import[TradeLead::Fbopen,no_purge]".  So basically it's just an extra command line option that gets passed around.  My intention was to find a way to do this so it could be easily replicated with other endpoints and wouldn't require splitting up the Fbopen importer

Gemfile
```diff
@@ -19,6 +19,7 @@ gem 'htmlentities'
 gem 'us_states', git: 'git://github.com/GSA-OCSIT/us_states.git'
 gem 'git'
 gem 'rdiscount'
+gem 'sidekiq'
```
calumhalcrow: You've added sidekiq to the Gemfile, but it seems you're not using it anywhere. In addition, in order to use sidekiq in staging/prod we need to make operational adjustments. Did you mean to add this here?

calumhalcrow:
@tmhammer I have looked over your code. The details of the full import look good. Looking at things from a higher level, I suggest you make the following changes:

1. Modify ita:import so that it accepts either a module or _an importer class_, rather than a model class.

2. Subclass the importer into two: one for the full import and one for the “patch” import. You’d have: TradeLead::Fbopen::FullData and TradeLead::Fbopen::PatchData. (I suggest this for [single responsibility](http://en.wikipedia.org/wiki/Single_responsibility_principle) reasons.)

3. Move lib/tasks/ita.rake#importers into a mixin that you can include in our modules, so that you can do `args.module_or_model_class.constantize.importers.each` in ita:import. Then for the TradeEvent module, rather than include the mixin, define the method explicitly, listing all the importers _except_ TradeLead::Fbopen::PatchData. This means that if you were to run `rake ita:import[TradeLead]`, a full import (from scratch) would be made. The “patch” import would be skipped.

4. Overwrite `TradeLead::Fbopen::PatchData#can_purge_old?` so that it always returns false.

5. Adjust [webservices-cookbooks](https://github.com/GovWizely/webservices-cookbooks/blob/master/gw_webservices/recipes/crontabs.rb) so that the crontab runs `rake ita:import[TradeLead]` every (say) Monday, and runs (something like) `rake ita:import[TradeLead::AustraliaData] ita:import[TradeLead::CanadaData]…[all importers except “FullData”]` on every other day of the week. That way, you’ll do a full import every Monday, and a patch every other day. The reason for point 4. above is so that when the rake task does `import_and_if_possible_purge_old`, it won’t purge old docs.

The "every other day" rake command in 5. may seem long, but I think it's good that way. We're doing a non-standard import, and so we're explicitly stating which importers should be run.

lib/tasks/ita.rake
```diff
@@ -23,19 +23,12 @@ namespace :ita do
...
     else
-      modu1e = module_or_model_class
-      module_importer_files =
-        "#{Rails.root}/app/importers/#{modu1e.name.typeize}/*"
-      Dir[module_importer_files].each { |f| require f }
-
-      modu1e.constants
-        .map { |constant| modu1e.const_get(constant) }
-        .select { |klass| klass.include?(Importer) }
+      module_or_importer_class.importers
```
calumhalcrow: @tmhammer you don't need to extend all importer classes with the `Importers` module; the above will do the job.

spec/importers/trade_lead/fbopen_full_data_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  let(:fixtures_dir) { "#{Rails.root}/spec/fixtures/trade_leads/fbopen" }
+  let(:resource) { "#{fixtures_dir}/full_source_input.xml" }
+  let(:importer)     { described_class.new(resource) }
+
+  it_behaves_like 'an importer which can purge old documents'
+
+  describe '#import' do
+    it 'loads leads from full xml' do
+      expect(TradeLead::Fbopen).to receive(:index) do |fbo|
+        expect(fbo.size).to eq(3)
```
calumhalcrow: Should you not be checking expected results here?

calumhalcrow:
With the addition of testing for expected results to TradeLead::FbopenImporter::FullData, I think this looks good. I'm currently running a full import locally, which is taking some time. I will deploy this to staging tomorrow my time (Tue evening EST).

Before merging I will check/do a couple of other things:
- that multiple downloads of the > 1GB source file don't remain in Rails' temp dir over long periods of time, using up disk space.
- Change [webservices-cookbooks](https://github.com/GovWizely/webservices-cookbooks) so that it does a full import once per week and a patch import once a day.

tmhammer:
Thanks @calumhalcrow, good call on the Importers.  Specs are updated and I also made some changes to the "end_date" field in the importers.

PR #104, opened 2015-03-09
Update importers, add Isa and Part561 sources

tmhammer:


calumhalcrow:
The change in this PR breaks out of the conceptual model that we've built up around consolidated endpoints. We have been mapping Sources, Importers and Models onto each other one-to-one. Now we have five "sources" all originating from the same...[source](https://www.youtube.com/watch?v=Ycu434kkk14)?

I considered suggesting that we adjust our conceptual model to cope with this. That we have only one Model class for the entire ScreeningList module (and all other modules for that matter), that we have one importer to handle the Treasury's consolidated list, that we set the document's "source" field accordingly, etc., but I think doing all that is outwith the scope of this change, especially given the time constraints.

So, I'm ok with these code changes. Note however that I changed the spec fixtures so that all importers which import from the consolidated source use the same consolidated file. We shouldn't test against individual files if we don't do that in production.

PR #105, opened 2015-03-13
Return error message for bad params

tmhammer:
[#87655654](https://www.pivotaltracker.com/story/show/87655654)


calumhalcrow:
This looks good. I think we should only apply it to calls to v2 endpoints though, as if we apply it to v1 we are changing the behavior of the endpoint and may break some peoples' integrations. I'm introducing a v2 controller in [#86679256](https://www.pivotaltracker.com/story/show/86679256) though, which would make this change simple to do, so maybe we can merge this PR after #86679256's has been merged?

calumhalcrow:
Also @tmhammer remember to run `rubocop -a`.

PR #106, opened 2015-03-20
Query string query for date ranges

tmhammer:
More dates, fixed specs

Rename 'faqs' to 'ita_faqs'


app/controllers/api/v1/parature_faq_controller.rb
```diff
@@ -1,4 +1,4 @@
 class Api::V1::ParatureFaqController < ApplicationController
   include Searchable
-  search_by :question, :answer, :update_date_start, :update_date_end, :countries, :industries, :q, :topics
+  search_by :question, :answer, :update_date, :countries, :industries, :q, :topics
```
calumhalcrow: @tmhammer removing the "update_date_start" and "update_date_end" params from this v1 endpoint will break the interface. I don't think we can do that. @ErikSArnold ?

config/routes.rb
```diff
@@ -25,7 +25,7 @@
   concern :api_routable do
 
     path = { 'market_researches'         => 'market_research_library',
-             'parature_faq'              => 'faqs',
+             'parature_faq'              => 'ita_faqs',
```
calumhalcrow: This will also break the v1 API, @ErikSArnold FYI.

calumhalcrow:
@tmhammer "dump.rdb" is included here. Should it be?

lib/query.rb
```diff
@@ -98,6 +98,14 @@ def generate_filter(json)
     filter_from_fields(json, query_fields)
   end
 
+  def query_string_query(json, value, field)
```
calumhalcrow: Can you call this `generate_query_string_query`? That would be more in line with other methods.

spec/fixtures/parature_faqs/filter_date.json
```diff
@@ -1,16 +1,13 @@
...
     "bool": {
       "must": [
         {
-          "range": {
-            "update_date": {
-              "from": "2013-01-11",
-              "to": "2013-03-18"
-            }
+          "query_string": {
+            "query": "update_date:[2013-01-11 TO 2013-03-18]"
```
calumhalcrow: I'm curious as to if/why the "query_string" approach is better than the "range" approach. Any thoughts on that @tmhammer?

spec/api/v2/screening_lists/consolidated_spec.rb
```diff
@@ -308,5 +308,33 @@
       end
     end
 
+    context 'when expiration_date is specified' do
+      subject { response }
+      let(:params) { { sources: 'SDN', expiration_date: '[* TO 2011-01-23]' } }
```
calumhalcrow: We need to be careful about what we allow to pass through here. See the warning in the [wildcard section](http://www.elastic.co/guide/en/elasticsearch/reference/1.x/query-dsl-query-string-query.html#_wildcards) of ES's documentation. If users can enter wildcards and we pass that on directly to the DB, they may put unacceptable load on the DB. What about ES's equivalent of [SQL injection](http://en.wikipedia.org/wiki/SQL_injection)?

calumhalcrow: BTW, a suitable solution may be to:

- keep the ES-like range syntax in the query string.
- break this into its component start and end parts in the query (raising an error if the format is invalid), then turning into a filter->range query for ES.

That way, you solve both issues I've raised about this (turning filter into query is bad for performance, not validated input from outside is dangerous).

tmhammer:
@calumhalcrow , For the V1 stuff, the Faqs were never being used on V1 so I don't think that will be an issue.  I think your proposed solution for using the range filter instead of query string sounds good so I'll work on making those changes.

tmhammer:
@calumhalcrow I think this is ready for another look.  I made a minor change to the Fbopen full import spec (didn't realize you had already merged a fix for this), where the importer wasn't importing one of the fixture entries because it had just become out of date.  I think it's handled now though.

calumhalcrow:
Switching to the branch that has this work based from "i-vs-i".

PR #107, opened 2015-03-22
Carlos/industries vs industry

carloslima:


calumhalcrow:
Thanks @carloslima. I just took a quick look over this and it all looks roughly ok. I will take a closer look today also, and add any changes to help get it ready that I can. I just rebased against master, so you'll have to reset before doing any further work. I see one failing spec after the rebase, but I'll fix that if I can.

PR #113, opened 2015-03-27
Makes dates searchable for appropriate data sets

calumhalcrow:
[#79898306]


calumhalcrow:
@tmhammer there's a couple of problems with this, both regarding invalid query params. Take a look at the following two requests:
```
$ curl "http://api.govwizely.com/v2/market_research_library/search?expiration_date=2017-11-30&api_key=-CO_3wIZfnBUOiWEjsLxcRiY" -v
...
< HTTP/1.1 500 Internal Server Error
```
This create an Airbrake ticket: https://govwizely.airbrake.io/projects/106408/groups/1385282394931234447/notices/1385282394931234446

```
$ curl "http://api.govwizely.com/v2/market_research_library/search?expiration_date=2017-11-30%20TO%202017-11-31&api_key=-CO_3wIZfnBUOiWEjsLxcRiY" -v
...
< HTTP/1.1 500 Internal Server Error
```
Created: https://govwizely.airbrake.io/projects/106408/groups/1385281124686268031/notices/1385281124686268030

--

We'd need these types of request to respond with a "400 bad request", with a relevant error message and without creating an Airbrake ticket. Right now they are responding with a 500, no error message and creating Airbrake tickets.

tmhammer:
@calumhalcrow , changed this so that passing in an invalid date range will now return a 400 bad request response instead of just raising an exception.  @ErikSArnold @jamesisaacs

PR #114, opened 2015-03-27
Remove individual endpoints from consolidated sources only for v2

baraabourghli:


calumhalcrow:
Great start @baraabourghli :-) I'm glad to see that you've been able to get started and make some actual changes. Your changes here are correct, you've understood the intent of the story. There's a couple of further changes I'd like you to make before this can be considered complete:

1. Can you do the same for all other endpoints that consolidate several sources into one? The other ones are TradeEvents and TradeLeads. TariffRates is also a consolidated endpoint, but IIRC we didn't add individual endpoints when we did the original implementation. I don't think there are any others, but please do double-check the routes to see if there are.

2. We put the spec examples for these endpoints in shared_examples, since they were being used by more than one spec (specifically, they were being used by two specs: the consolidated one and the individual ones). Now that you're removing one of those two specs, can you remove the shared_examples and place the examples directly in the spec? The specs will become more straight-forward that way.

Let me know if you have any questions.

baraabourghli:
Thanks @calumhalcrow :)

I only found individual endpoints in TradeEvents, also the shared_examples are fixed.

calumhalcrow:
Cool, will check tomorrow morning.

calumhalcrow:
This is heading in the right direction.

@baraabourghli can you rebase onto master? There are some conflicted changes due to new code that has been commit to master recently.

I've been reviewing the specs in general today, and as part of your work on this I'd like to better organize the data files that support the specs. The way things are now is both confusing and inconsistent. Specifically I'd like to do the following:

1. Adjust names of request specs so that they correspond with API URLs.
2. Move expected results data files so that they live with their corresponding specs:
      - API responses with their specs in spec/api/vX/consolidate_screening_list/ etc.
      - expected indexed docs YAML in spec/importers/
      - expected query JSON in spec/queries/
      - leave only the example sources in spec/fixtures.

Tomorrow I will apply these changes to the "market research library" specs, to give you a concrete example of what I'd like. After that I think I'll create another story for this work, and assign it to you, rather than ask you to do it as part of this PR. The end goal here is to get our API endpoints sorted out, and also the specs that go with them.

If you can get the rebase done I'll double-check everything and then merge.

spec/support/screening_lists_search_behavior.rb
```diff
@@ -586,18 +412,6 @@
...
-  let(:expected) { (0..3).to_a }
-  it_behaves_like 'it contains all expected results of source'
-end
-
-shared_examples 'it contains all ScreeningList::Ssi results that match countries "UA,DJ"' do
-  let(:source) { ScreeningList::Ssi }
-  let(:expected) { [] }
-  it_behaves_like 'it contains all expected results of source'
-end
-
```
calumhalcrow: You've uncovered an interesting situation. All the shared_examples that you've removed from this file are not being used in either of the v1 and v2 consolidated specs, so removing them makes sense. Yet I believe they should in fact be in use in the consolidated specs! I'm going to attempt to add them to the v2 consolidated spec now. If that works out as expected, I'll merge the PR...

calumhalcrow: After reviewing the consolidated spec more closely, I can see that coverage of the various query params is already sufficient. So I'm going to merge this as it is now.

PR #120, opened 2015-04-01
Suppress unpermitted parameters errors

baraabourghli:
This is an initial solution, another way of doing it is adding

    config.action_controller.action_on_unpermitted_parameters = :log

to `config/environments/production.rb`, but I am not sure which solution is more suitable for our case.

### DIFF
--- a/app/controllers/api/v2_controller.rb
+++ b/app/controllers/api/v2_controller.rb
@@ -1,13 +1,6 @@
 class Api::V2Controller < ApiController
   before_action :authenticate_by_api_key
 
-  ActionController::Parameters.action_on_unpermitted_parameters = :raise
-
-  rescue_from(ActionController::UnpermittedParameters) do |e|
-    render json:   { error:  { unknown_parameters: e.params } },
-           status: :bad_request
-  end
-
   rescue_from(Exceptions::InvalidDateRangeFormat) do |_e|
     render json:   { error:  'Invalid Date Range Format' },
            status: :bad_request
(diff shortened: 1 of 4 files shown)


calumhalcrow:
@baraabourghli if we log then the server responds with 200, which is not the behavior we want. We want it to do what it already does: return 400 and an error message. If you look [here](https://github.com/GovWizely/webservices/blob/master/app/controllers/api/v2_controller.rb#L6) you'll see that we have code in place to deal with this.

The problem (and mystery as far as I'm concerned) is that Airbrake is aware of these errors and reports them to us. It doesn't do so for the `Exceptions::InvalidDateRangeFormat` error that's handled further down in the code I linked to. Airbrake telling us that this error was encountered is useless to us, so I'd like not to receive it.

I realize that we could just configure Airbrake so that it doesn't inform us. The part that I'd like to understand is why we get messages about `ActionController::UnpermittedParameters` but not `Exceptions::InvalidDateRangeFormat`. If you cannot figure it out then no worried, we can resort to configuring Airbrake if necessary.

baraabourghli:
@calumhalcrow if you look at the errors in Airbrake, it seems like they are all coming from V1, handling was done in V2 only.

calumhalcrow:
I see what you mean. Mystery solved.

calumhalcrow:
I'm adjusting the specs a bit, but I will merge this tonight (probably at Ruby Tuesday!)

baraabourghli:
Sounds good :+1:

PR #123, opened 2015-04-03
Fuzzy

calumhalcrow:


calumhalcrow:
@jamesisaacs I have opened this even though the work may not be ready for merge. I just want a place to be able to give feedback.

app/queries/screening_list/query.rb
```diff
@@ -16,6 +16,7 @@ def initialize(options = {})
       @start_date = options[:start_date] if options[:start_date].present?
       @issue_date = options[:issue_date] if options[:issue_date].present?
       @expiration_date = options[:expiration_date] if options[:expiration_date].present?
+      @phonetics = options[:phonetics] if options[:phonetics].present?
```
calumhalcrow: I feel you should do `@phonetics = true if options[:phonetics].present?`, so that later you have conditions like:
```
if @phonetics
  # phonetics stuff
end
```
rather than:
```
if @phonetics == '1'
  # phonetics stuff
end
```

spec/support/screening_lists_search_behavior.rb
```diff
@@ -366,7 +366,7 @@
 
 shared_examples 'it contains all ScreeningList::Plc results that match "SALEH, Jamal" with fuzziness of 1' do
   let(:source) { ScreeningList::Plc }
-  let(:expected) { [3] }
+  let(:expected) { [0, 3] }
```
calumhalcrow: It's a mystery to me as to why this isn't already an issue on master.

app/queries/screening_list/query.rb
```diff
@@ -30,25 +31,29 @@ def generate_query(json)
           end if @q
 
           if @name
-            generate_fuzzy_queries(json, %w(name.keyword alt_names.keyword), @name)
+            # generate_fuzzy_queries(json, %w(name.keyword alt_names.keyword), @name)
+            generate_fuzzy_queries(json, %w(name alt_names), @name)
+          end
+          if @phonetics == '1'
+            generate_fuzzy_queries(json, ['phonetic_names'], @name)
```
calumhalcrow: We don't have coverage for this line.

screen shot 2015-04-03 at 3 45 12 pm

app/models/screening_list/mappable.rb
```diff
@@ -35,6 +46,7 @@ def self.included(klass)
           properties: {
             name:                    { type:     'string',
                                        analyzer: 'standard_asciifolding_nostop',
+                                       boost:    4.0,
```
calumhalcrow: Like I mentioned, after reading [the docs](http://www.elastic.co/guide/en/elasticsearch/reference/1.x/mapping-boost-field.html) I don't think this is going to do anything. I think you're trying to tell ES to give the "name" field more weight when performing a query. When I read the docs, it seems that in order to tell it that, you need to do so via the _query_, no via the _mapping_. I could be wrong though, tell me if I am :-)

Ideally, it would be good to have a spec that demonstrates that our weighting works. For example a spec that shows that a match in the name field gets a higher score than a match in other fields. Something like:

```
# We have two docs. A query for "father xmas" returns them both. 1's score is higher than 2's:
{ id: 1, name: 'Father Xmas', alt_names: ['Santa Claus'] },
{ id: 2, name: 'Santa Claus', alt_names: ['Father Xmas'] }
```

app/queries/screening_list/query.rb
jamesisaacs: Good point. I've made this change.

jamesisaacs:
@calumhalcrow I've added `max_score` to the json results. We'll need to filter that out of the specs along with `score` as a short-term fix. Adding `max_score` also required me to either add `track_score` to the query, which comes at a processing time cost, or to remove `@sort` from `query.rb`. I removed the sort thinking the default was what we wanted anyway. Please  let me know if there's something I missed.

calumhalcrow:
@jamesisaacs spec/services/recreate_indices_with_modified_mappings_spec.rb is failing because the mapping defined in `ScreeningList::Mappable` is not saved to ES as-is. ES modifies it slightly during the save. We need the mapping defined in the code to match what ES saves, so that we know which indices have changed at deployment time.

It seems that ES disregards the [`boost: 1.0`](https://github.com/GovWizely/webservices/blob/james/fuzzy/app/models/screening_list/mappable.rb#L49). This is what I see in ES:

screen shot 2015-04-10 at 5 26 05 pm

Are you sure that adding the "boost" param is working as intended?

app/controllers/api/v1/screening_lists/consolidated_controller.rb
```diff
@@ -1,4 +1,4 @@
 class Api::V1::ScreeningLists::ConsolidatedController < ApiController
   include Searchable
-  search_by :countries, :q, :type, :sources, :name, :fuzziness, :address
+  search_by :countries, :q, :type, :sources, :name, :distance, :address
```
calumhalcrow: Generally, we try not to change the v1 interface, but maybe in this case the "fuzziness" feature isn't documented and hence isn't in active use..?

app/models/screening_list/mappable.rb
```diff
@@ -33,16 +34,84 @@ def self.included(klass)
...
 
+            name_idx:                { type:     'string',
+                                       analyzer: 'standard_asciifolding_nostop',
+                                       fields:   {
+                                         keyword: { type: 'string', analyzer: 'keyword_lowercase' } } },
+
+            name_no_common:          { type:     'string',
+                                       analyzer: 'standard_asciifolding_nostop',
+                                       fields:   {
+                                         keyword: { type: 'string', analyzer: 'keyword_lowercase' } } },
```
calumhalcrow: What comes to mind when I see this is the [multi field type](https://www.elastic.co/guide/en/elasticsearch/reference/0.90/mapping-multi-field-type.html). I imagined the mapping would look something like this:
```
             name:                    { type:     'string',
                                        analyzer: 'standard_asciifolding_nostop',
                                        fields:   {
                                          keyword: { type: 'string', analyzer: 'keyword_lowercase' },
                                          idx: { type: 'string', analyzer: 'CUSTOM IDX ANALYZER' },
                                          no_common: { type: 'string', analyzer: 'CUSTOM NO_COMMON ANALYZER' },
                                           ... etc. ...
                                        },
             },
```
Maybe it's not possible to fashion analyzers to these specific analyzes..?

PR #125, opened 2015-04-09
mca leads

baraabourghli:


baraabourghli:
@calumhalcrow The specs in V1 are not passing because we are importing to ES index the new MCA data, how can we avoid/fix this?

app/queries/v2/trade_lead/query.rb
```diff
@@ -13,6 +13,7 @@ def initialize(options = {})
       @publish_date = options[:publish_date] if options[:publish_date].present?
       @end_date = options[:end_date] if options[:end_date].present?
       @publish_date_amended = options[:publish_date_amended] if options[:publish_date_amended].present?
+      @categories  = options[:categories].split(',').map(&:strip) rescue nil
```
calumhalcrow: Even though you'll see `var = statement rescue nil` here and there in the codebase, it's [not good to do so](https://github.com/bbatsov/ruby-style-guide#no-rescue-modifiers). Better to do `var = statement if something`.

calumhalcrow:
Try running `rubocop -a`. It will correct some styles issues.

app/views/api/v2/trade_leads/consolidated/search.json.jbuilder
```diff
@@ -18,6 +18,7 @@ field_lists = {
   uk:        [:country, :reference_number, :publish_date, :min_contract_value, :max_contract_value,
               :status, :url, :procurement_organization, :contact, :title, :description, :notice_type,
               :industry, :specific_location, :source],
+  mca:        [:country, :publish_date, :url, :category, :title, :description, :category, :funding_source, :source],
```
calumhalcrow: You've got `:category` in there twice.

calumhalcrow:
@baraabourghli regarding your question about MCA results showing up in v1: the codebase doesn't have a convenient way of searching different indices depending on the API version. There are three solutions that come to mind:

1. Add the source to v1.
2. Allow ES to search over the MCA index, but (via `Api::V1::TradeLeads::ConsolidatedController` and its templates) don't show the documents in the results.
3. Adjust the codebase so that it can deal with different sets of sources for different API versions.

I don't like 3. because I think it's too much work to be worthwhile. 1. would be easiest from a coding point of view, so I'll ask Stuart if it's ok to do that. If 1. is a no-go, we'll do 2.

app/controllers/api/v2/trade_leads/consolidated_controller.rb
```diff
@@ -1,4 +1,4 @@
 class Api::V2::TradeLeads::ConsolidatedController < Api::V2Controller
   include Searchable
-  search_by :countries, :industries, :q, :sources, :publish_date, :end_date, :publish_date_amended
+  search_by :countries, :industries, :q, :sources, :publish_date, :end_date, :publish_date_amended, :categories
```
calumhalcrow: I'm not sure if we'll actually need to support a query param for searching categories. I'd like to see the Google doc first though (access pending).

calumhalcrow:
@baraabourghli just for completeness, can you access CodeClimate: https://codeclimate.com/github/GovWizely/webservices/compare/mca-leads ?

baraabourghli:
@calumhalcrow yes I can access CodeClimate.

calumhalcrow:
:+1:

spec/api/v1/trade_leads_spec.rb
```diff
@@ -62,9 +63,9 @@
         expect(json_response['offset']).to eq(0)
 
         results = json_response['results']
-        expect(results[0]).to eq(expected_results[1])
-        expect(results[1]).to eq(expected_results[2])
-        expect(results[2]).to eq(expected_results[3])
+        expect(results[0]).to eq(expected_results[4])
```
baraabourghli: For the future, what do you think of having atomic specs like:
    
    it { expect (results[0]).to eq expected_results[4] }

This might help identifying exactly what failed instead of debugging!?

spec/api/v1/trade_leads_spec.rb
```diff
@@ -25,7 +26,7 @@
 
       it 'returns trade leads sorted by published_date:desc, country: asc' do
         json_response = JSON.parse(response.body)
-        expect(json_response['total']).to eq(13)
+        expect(json_response['total']).to eq(16)
```
baraabourghli: Maybe later on we can divide the `expected_results_v1.json` file into multiple files, that way it will be less confusing why we expect the `total` to be `16` but then we match `expected_results.first(10)`, what do you think?

baraabourghli:
@calumhalcrow for V1 support, I only added it to consolidated search, is this okay?

baraabourghli:
@calumhalcrow it just failed on my local, I believe that the index is getting import twice, I am not sure how to avoid that, so what do you suggest?

This is what i am getting in my local:

    Failure/Error: expect(json_response['total']).to eq(16)

       expected: 16
            got: 19

       (compared using ==)
     # ./spec/api/v1/trade_leads_spec.rb:45:in `block (4 levels) in <top (required)>'

spec/api/v1/trade_leads_spec.rb
calumhalcrow: For sure :-) In fact, I think we could move most of the detailed search testing into the Query specs, and just have these "api" (i.e. request) specs do a few sanity checks instead. Better to have all the edge case testing close to the implementation, and do fewer high-level tests.

spec/api/v1/trade_leads_spec.rb
calumhalcrow: I think this is a side-effect of the fact that we test most of our search functionality through the request specs rather than the Query specs (like I mentioned earlier). I'd rather we move detailed testing of search into Query specs, then just have less documents imported for request specs, so that we don't even need to think about problems like this. But that's a refactoring for another day, not part of this story.

calumhalcrow:
Regarding "only adding to consolidated": yeah that's fine.

calumhalcrow:
Regarding importing twice: yes, it probably is. The best thing to do is to ensure your importer assigns an id to the doc. If ES indexes a doc which has an id, and an existing doc has the same id, it will overwrite rather than create. Follow the example here: https://github.com/GovWizely/webservices/blob/master/app/importers/trade_event/dl_data.rb#L40

app/controllers/api/v2/trade_leads/consolidated_controller.rb
calumhalcrow: Yeah, there's no mention of adding this param, so please remove :-)

app/importers/trade_lead/mca_data.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    private
+
+    def process_entry(item)
+      item_hash = extract_fields(item, SINGLE_VALUED_XPATHS)
+      item_hash.reverse_merge! extract_multi_valued_fields(item, MULTI_VALUED_XPATHS)
+      item_hash[:url].gsub!(/www\./, 'mcc.')
+      country = item_hash[:category].delete_at(item_hash[:category].find_index { |e| /country\// =~ e })
+      country = country.match(/country\/(\w\w)/)
+      item_hash[:country] = country[1] if country.present?
```
calumhalcrow: Can you make sure the 2 character country code is upper-cased, and also that it is a valid ISO code. You can use the gem that's used here: https://github.com/GovWizely/webservices/blob/master/lib/importer.rb#L37

baraabourghli: Sure :+1:

app/controllers/api/v2/trade_leads/consolidated_controller.rb
baraabourghli: Alright, I will remove it then.

lib/utils.rb
```diff
@@ -0,0 +1,7 @@
+module Utils
+  module_function
+
+  def generate_id(hash = {}, keys = [])
+    Digest::SHA1.hexdigest(keys.map { |k| hash[k] }.join)
+  end
+end
```
calumhalcrow: Nice use of `module_function`.

Also, I see you still like to use Sublime ;-)

baraabourghli: Actually I joined the Rubymine tripe recently, but this one escaped the checking somehow :-P

calumhalcrow: Oh dear... :D

baraabourghli: Fixed.

PR #132, opened 2015-04-17
Spec data reorg

calumhalcrow:


calumhalcrow:
@baraabourghli looks ok. The approach is:
- only have _input_ (i.e. source) files in /spec/fixtures
- all other data files should be as close to the relevant spec as possible (in the same dir or in a child dir with the same name).

baraabourghli:
@calumhalcrow BTW, I wanted to do specs per source, but I was worried about making huge changes per commit and missing some details. Anyway, I can rewrite the commits history and organize changes per source if you think it is better to have?

baraabourghli:
@calumhalcrow I think it is ready to be reviewed.

calumhalcrow:
@baraabourghli I don't think it matters about how the changes are chopped up into commits. This looks good. I'll do some local testing, then merge and deploy to staging.

PR #139, opened 2015-04-29
Record failed lookups only in production.

baraabourghli:


baraabourghli:
Blocked by https://github.com/GovWizely/industry_mapping/pull/4

PR #140, opened 2015-04-29
Negative offset handling 

baraabourghli:
I want to verify if you feel comfortable adding Rack middlewares!
P.S, currently tests will not cover the middleware, we need to add integration tests https://stackoverflow.com/questions/5016690/making-rails-tests-aware-of-rack-middleware-outside-railss-internal-chain

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -6,6 +6,11 @@ class ApiController < ActionController::Base
            status: :bad_request
   end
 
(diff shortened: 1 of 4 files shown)


calumhalcrow:
I don't mind middleware, but in this case I think it's better to put a validation on the appropriate object. In this case, it's the Query, since "offset" it an attribute of it. I've added a commit showing what I mean. Can you work from there and add specs etc?

baraabourghli:
Got it :+1:

PR #147, opened 2015-05-21
Support responsive form input fields

baraabourghli:


app/assets/stylesheets/customizations.css.scss
```diff
@@ -36,6 +36,10 @@
           float: left;
           width: 12rem;
           line-height: 1em;
+          /* Smartphones (portrait and landscape) ----------- */
```
baraabourghli: Source: http://responsivedesign.is/develop/browser-feature-support/media-queries-for-common-device-breakpoints

calumhalcrow:
I adjusted the change so that the media query is kept separate from the rest of the styles (which is conventional), but apart from that looks good. I'll deploy it to staging so that we can test on iphones etc.

PR #148, opened 2015-05-21
Obtain BSP XML from Emenu

joelanders:
Updated:

I did my best guesses at the indexing settings--maybe they should be different, though.
There's some matter-of-taste stuff with the route path and the parameter names.
And maybe you like the *Query spec a different way.

But other than that, it works.


calumhalcrow:
Some feedback:

- There are some recent changes on master that you can pick up with a `git rebase origin/master`.
- I'm getting a "missing template" error when I try to load the v2 URL (http://localhost:3000/v2/emenu_bsps/search).
- Run `rubocop -a

spec/importers/emenu_bsp_data_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+describe EmenuBspData do
+  let(:fixtures_dir) { "#{File.dirname(__FILE__)}/emenu_bsp" }
+  let(:fixtures_file) { "#{Rails.root}/spec/fixtures/emenu_bsp_articles/emenu_bsp_articles.json" }
+  let(:importer) { EmenuBspData.new(fixtures_file) }
+  let(:articles_hash) { YAML.load_file("#{fixtures_dir}/emenu_bsp_articles.yaml") }
+
+  it_behaves_like 'an importer which cannot purge old documents'
```
calumhalcrow: This importer should be purging old docs.

joelanders:
Ok, I added the views first.
re: rubocop, should I commit all of the suggested changes, or just the ones relevant to this branch?
`rubocop $(gd f2e2a312bf61e692b7f49303ea80a32b56f624c1 --name-only)`

calumhalcrow:
@joelanders for rubocop, just do `rubocop -a` and it will auto-fix anything that it can. You should find that if you run `rubocop -a` on master there will be no resulting changes, so by a global run on your branch will only adjust your work. If you want to then look over the advice that can only be done manually, by all means act on anything that you can, but doing so isn't compulsory, and you will find problems that are not related to your work.

joelanders:
Re: `filter_from_fields`: it depends on  `query_fields` and it looks like `setup_query` is the only thing that sets that. And I think that method is only called in `ItaOfficeLocationQuery` and `EmenuBspQuery`.

app/importers/emenu_bsp_data.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  private
+
+  def process_article_info(article_hash)
+    article = remap_keys COLUMN_HASH, article_hash['company']
+    article[:ita_office] = article_hash['site']
+    article[:contact_title] = article_hash['company']['contact']['title']
+    article[:contact_name] = article_hash['company']['contact']['name']
+    category = article_hash['company']['category']['title']
+    article[:category] = CATEGORIES.include?(category) ? category : nil
```
calumhalcrow: I'm curious about this requirement. Is it mentioned somewhere that the category field must be included in this list?

joelanders: Ah, yeah the Google Doc said this one had a "controlled vocabulary" and provided this list. Maybe that just meant to use keyword analysis in ES?

PR #154, opened 2015-06-02
Joe/resource versioning

joelanders:


joelanders:
Ah, I've got to rebase this to hit `app/models/concerns/{indexable,searchable}.rb` instead of `lib{indexable,searchable}`.

joelanders:
@nickmarden https://github.com/GovWizely/webservices/commit/96e9eb5600e4da7d0911f8580c3bfb21f6f13f13

That shared concern just checks the child class concretely implements `available_version`. This spec^ checks that if `available_version` is implemented, the `Importer` code does the right thing.

app/models/concerns/indexable.rb
```diff
@@ -30,6 +30,21 @@ def create_index
...
+        type: 'metadata',
+        id: 0,
+        body: { version: version, time: DateTime.now.utc })
+    end
+
+    def stored_metadata
+      ES.client.search(
+        index: index_name,
+        type: 'metadata',
+        body: {query: {match: {_id: 0}}})['hits']['hits'][0]['_source'].symbolize_keys rescue {}
```
calumhalcrow: BTW, suppressing exceptions like this is [considered bad form](https://github.com/bbatsov/ruby-style-guide#dont-hide-exceptions). You need to ask yourself: under what circumstances do I expect this statement to fail, and can I rewrite it to avoid the failure? If not, you should be more explicit about the type of exception you expect to catch.

lib/importer.rb
```diff
@@ -8,18 +8,34 @@ def self.included(base)
...
 
-      start_time = Time.now if can_purge_old?
-      super
-      model_class.purge_old(start_time) if can_purge_old?
+      if resource_changed?
+        start_time = Time.now if can_purge_old?
+        super
+        update_metadata available_version
+        model_class.purge_old(start_time) if can_purge_old?
+      end
```
calumhalcrow: Can you log the fact that the import was skipped due to the resource being unchanged?

app/models/concerns/indexable.rb
joelanders: Yeah I guess what I want is something like short-circuit evaluation on a chain of method calls: `['hits']['hits'][0]['_source']` -- instead of `if/else` checking for all of those keys, just let me give another expression for when one of them is absent.
I wish you could do `{}[0][0] rescue(NoMethodError) {}`.
I also wish you could just catch exceptions thrown in the immediate execution context: `ES.client.search(...)[:some][:keys] rescue_local(NoMethodError) {}` but I guess that's really being thrown in an instance of NilClass...
    class << nil
      def [](*_); self; end
    end
Hmm... screw it.

calumhalcrow: It's not critical that you fix this. I just noticed you using the `some statement rescue default value` pattern a lot and wanted to highlight the fact that it's not best practice. If you don't change this I'll still merge the PR.

calumhalcrow:
@joelanders can I remove the "do not merge" label?

calumhalcrow:
I ran most importers locally to check that they still work. I got an error when importing TradeEvent::UstdaData:

```
Encoding::UndefinedConversionError: "\xD0" to UTF-8 in conversion from ASCII-8BIT to UTF-8 to UTF-16LE
```

The same importer works on master. You can run the importer with `rake ita:import[TradeEvent::UstdaData]` and see the result in the sidekiq UI at http://localhost:3000/sidekiq

calumhalcrow:
You need to rebase this also.

calumhalcrow:
Ok, this looks look. Once you've fixed the UstdaData problem I'll deploy this to staging, and let it run there for a few days so that I can verify that all imports are working ok. Then I'll merge. Good work here :+1:

joelanders:
Forced the encoding and did a rebase.

calumhalcrow:
This is on staging. I'll merge to master once I'm happy that all importers are working as expected there.

PR #157, opened 2015-06-08
Routes and CC update

tmhammer:


calumhalcrow:
Awesome Tim! You've fixed the routing problems, and also fixed every issue that was raised by CC :+1:

PR #159, opened 2015-06-12
Elasticsearch in Procfile

calumhalcrow:
@carloslima @jamesisaacs would you be so kind as you give your opinions on this change?

I'm trying to kill a couple of birds with this PR:
- In dev envs, tie the running of ES to foreman, so that it stops when everything else stops. It tends to kill my laptop battery otherwise (thanks Java), and I'm forever starting/stopping it independently.
- Move db (i.e. index) creation into a rake task, making it more rails-like and (more importantly) putting an end to checking if indices exist every time a rails process starts. We don't need that, especially in staging/prod. Whenever we want to drop/create all indices let's use rake tasks, just like most other rails apps.


jamesisaacs:
@calumhalcrow Both changes sound good to me.

calumhalcrow:
@jamesisaacs thanks. I'll give @carloslima a day or so to respond if he likes, after which I'll assign to you for merge.

calumhalcrow:
Carlos say :+1:

PR #160, opened 2015-06-15
Bugfix for rev w/ common and refactoring

jamesisaacs:
The bug fix can be seen clearly in this commit: https://github.com/GovWizely/webservices/commit/7ab60e5705100df4b23fce41977fbcc8d7b9a3e2 Some alternative names with common words weren't being reversed properly.


app/importers/screening_list/make_name_variants.rb
```diff
@@ -12,45 +12,55 @@ module MakeNameVariants
...
+      name.class == String ? name.split.delete_if { |n| wordlist.include?(n.downcase) }.join(' ') : name.map { |n| n.split.delete_if { |word| wordlist.include?(word.downcase) }.join(' ') }
+    end
+
+    def name_rev(name)
+      name.class == String ? name.split.reverse.join(' ') : name.map { |n| n.split.reverse.join(' ') }
+    end
+
+    def name_trim(name)
+      name.class == String ? name.gsub(/\s+/, '') : name.map { |n| n.gsub(/\s+/, '') }
     end
```
calumhalcrow: Ideally you'd have a spec for this module which showed these methods in action against simple examples. It's especially useful to create such specs when a bug is found, as you're able to show "before" (failing) and "after" (passing) behaviors.

PR #161, opened 2015-06-15
Est endpoint

baraabourghli:


app/importers/est_data.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  private
+
+  def fetch_data
+    headless_login
+    mechanize_agent.get(@resource).body
+  end
+
+  def headless_login
```
baraabourghli: i couldn't find a way to test this function in Rspec without the access tokens.. is there some way to do it?

app/models/est.rb
```diff
@@ -0,0 +1,35 @@
+class Est
+  include Indexable
+
+  self.settings = {
+    index: {
+      analysis: {
+        analyzer:
+            { custom_analyzer:        {
+              tokenizer: 'standard',
+              filter:    %w(standard asciifolding lowercase snowball) },
```
baraabourghli: I am not sure if we should use `asciifolding` & `snowball` with portuguese, chinese, and spanish.

app/models/est.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  self.mappings = {
+    name.typeize => {
+      _timestamp: {
+        enabled: true,
+        store:   true,
+      },
+      dynamic:    'false',
+      properties: {
+        name_chinese:      { type: 'string', analyzer: 'custom_analyzer' },
```
baraabourghli: should i analyze all ?

app/importers/est_data.rb
```diff
@@ -0,0 +1,62 @@
+class EstData
+  include Importer
+  LOGIN_URL = 'https://admin.export.gov/site_login'
+  ENDPOINT = 'https://admin.export.gov/admin/envirotech_solutions.json'
+
+  cattr_accessor :web_auth, :username, :password
+
+  COLUMN_HASH = {
```
baraabourghli: is those names enough?

calumhalcrow: We don't have any guidance on this so I suggest indexing all available fields, snake_casing their names. If that's what you've done, then that's enough.

app/models/est.rb
calumhalcrow: It's worth thinking about for sure :+1: 

[asciifolding](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-asciifolding-tokenfilter.html) would allow our users to find a document with e.g. the name "René" by searching for "Rene", which is desired behavior (note that this is relevant to when a user does a search; the actual document content won't be altered by this filter).

The [snowball _filter_](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-snowball-tokenfilter.html) is a stemming algorithm that is language dependent. You can specify which language is being analyzed, so the most correct thing to do here would be to create a filter for each language, and analyze each field depending on which language it is in, like as shown in the snowball filter documentation. I recommend you do that (yes, there will be an analyzer for each language).

app/models/est.rb
calumhalcrow: As mentioned in my [earlier comment](https://github.com/GovWizely/webservices/pull/161/files#r32501928), yes please analyze all these by creating an analyzer for each language. snowball isn't suitable for Chinese though, but maybe just the standard word-based filter is good enough.

It will be worth adding specs to show some of these stemming/folding behavior in action :wink:

app/importers/est_data.rb
calumhalcrow: Don't worry about testing this. The way that you have mocked `fetch_data` in the specs is appropriate. Since all it does is communiate over the internet with the external service holding the source docs, we don't need to provide coverage of it. We can add `# :nocov:` comments around it to tell simplecov not to complain that it hasn't been covered.

baraabourghli:
Fixed.

calumhalcrow:
That looks about right. I'll wait for feedback on the endpoint URL before looking at this in more detail tomorrow...

calumhalcrow:
@baraabourghli as per Stuart's feedback [on the story](https://www.pivotaltracker.com/n/projects/1085558/stories/95859342), please change the endpoint URL to `/environmental_solutions/search`.

In order to keep naming consistent, I suggest replacing the name "est" with "environmental_solutions" throughout the codebase also. I know it's a bit of a pain to do so, but it will be better in the long run. E.g., the model would become `EnvironmentalSolution`.

calumhalcrow:
@baraabourghli there are actually more than 30 documents. I accidentally hid this fact by putting a `.json` link in the story's description. If you visit https://admin.export.gov/admin/envirotech_solutions you'll see that there are four pages of documents. We need them all.

app/importers/est_data.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def initialize(resource = ENDPOINT)
+    @resource = resource
+  end
+
+  def import
+    data = JSON.parse(fetch_data)
+    articles = data.map { |article_hash| process_article_info article_hash }
+
+    Est.index articles
```
calumhalcrow: You can do `model_class.index(articles)` here.

calumhalcrow:
Make sure you run these changes through all of our code quality tools: CodeClimate, rubocop, etc.

spec/api/v1/est_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      it 'returns the result matching with stemming/folding' do
+        json_response = JSON.parse(response.body)
+        expect(json_response['total']).to eq(1)
+        expect(json_response['offset']).to eq(0)
+
+        results = json_response['results']
+        expect(results.first).to eq(one_match.first)
+      end
+    end
+  end
```
calumhalcrow: Can you move the stemming example to the v2 spec? At some point all the v1 code will be deleted, so it would be better to have the bulk of the testing in v2. Also, can you add some more examples, like Chinese character matching?

app/importers/est_data.rb
baraabourghli: Yep :+1:

config/initializers/est.rb
```diff
@@ -0,0 +1,3 @@
+EstData.username = 'erik@govwizely.com'
```
baraabourghli: Is this intentional?

baraabourghli:
CodeClimate: https://codeclimate.com/github/GovWizely/webservices/compare/EST seems okay, unless you think those issues worth improving, then I will invest more time on improving them.

calumhalcrow:
@baraabourghli regarding CodeClimate: as long as the score doesn't go down, we're good :+1:

baraabourghli:
Okay then, I removed "do not merge" label.

config/est.rb
```diff
@@ -0,0 +1,3 @@
+EstData.username = '-'
+EstData.password = '-'
+EstData.web_auth = '-'
```
calumhalcrow: Note: we need to remove this file (probably my fault that it's still here).

calumhalcrow:
@baraabourghli can you `git rebase origin/master` this then force push?

baraabourghli:
Done @calumhalcrow.

calumhalcrow:
@baraabourghli this looks good. I'll wait for feedback from Stuart before merging though.

PR #167, opened 2015-06-18
ECCN source changed to TSV.

calumhalcrow:


jamesisaacs:
@calumhalcrow Stuart had a look and has approved deploying this to production. That will be done momentarily.

calumhalcrow:
Unfortunately an earlier version of this is already live on production. I'm going to introduce a staging rails environment, then remove the route to this endpoint `if Rails.env.production?` so that we can satisfy Stuart's requirement. As such, I might merge this PR today (if I am able to put the staging env changes in place).

PR #169, opened 2015-06-22
Carlos/fix fbopen full importer

calumhalcrow:


calumhalcrow:
@carloslima what's the root cause of the bug?

app/importers/trade_lead/fbopen_importer/full_data.rb
```diff
@@ -62,7 +72,7 @@ def process_xml_entry(entry)
         return nil if !(entry[:specific_location].try(:upcase) =~ /\A[A-Z]{2}\z/) || entry[:specific_location] == 'US'
 
         entry = process_xml_dates(entry)
-        return nil unless entry[:end_date] >= Date.today || entry[:end_date].nil?
+        return nil unless entry[:end_date].nil? || entry[:end_date] >= Date.today
```
carloslima: @calumhalcrow This ^^
Wrong ordering.

calumhalcrow: Heh, yeah that would be a problem if `entry[:end_date]` was nil :laughing:

calumhalcrow:
You've imporved our CodeClimate score by 0.06 :+1: 

https://codeclimate.com/github/GovWizely/webservices/compare/carlos/fix-fbopen-full-importer

calumhalcrow:
Coverage is < 100% though. The case where `entries.size >= 1000` isn't covered.

carloslima:
Should be better now :-)

PR #170, opened 2015-06-28
Carlos/fix empty q

carloslima:
if the "q" param is present but empty, the endpoints should behave as if no "q" parameter was present.

### DIFF
--- a/lib/query.rb
+++ b/lib/query.rb
@@ -35,7 +35,7 @@ class Query
 
     @offset = options[:offset].to_i
     @size   = [options[:size].to_i, MAX_SIZE].min
-    @q      = options[:q]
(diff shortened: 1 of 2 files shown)


calumhalcrow:
I'm still seeing the problem locally with this change in place. Can you verify locally that the following two URLs return the same data (after an import has been done of course :smiley:):

http://localhost:3000/market_research_library/search
http://localhost:3000/market_research_library/search?q=

calumhalcrow:
Also, better rebase :kissing_heart:

carloslima:
Ouch. :sweat_smile: 
Should be fixed now :white_check_mark:  :smile_cat:

PR #174, opened 2015-07-05
Get all MRL documents

baraabourghli:


spec/api/v2/market_research_library/search_spec.rb
```diff
@@ -134,4 +134,18 @@
       end
     end
   end
+
+  describe 'GET /market_research_library/search.json?size=-1' do
```
baraabourghli: I am not sure if we need to test the ES scroll by adding 100 documents.
If you think it is needed, please guide me on how to generate something similar to https://github.com/GovWizely/webservices/blob/master/spec/fixtures/market_research/source.txt

app/models/concerns/searchable.rb
```diff
@@ -71,17 +71,22 @@ def index_names(sources = nil)
...
-        batch = response['hits']['hits'].map(&:deep_symbolize_keys)
-        results.push(*batch)
+        case format
+          when :json
+            batch = response['hits'].deep_symbolize_keys
+            results.merge!(batch)
+          else
+            batch = response['hits']['hits'].map(&:deep_symbolize_keys)
+            results.push(*batch)
+        end
```
calumhalcrow: Here's how I see this code: `fetch_all` is responsible for fetching all documents from an index. It doesn't care about the requested output format, only that the caller wants all documents. Passing the requested format is polluting the simple nature of the method with information it shouldn't need.

But you are passing it for a reason. The reason is that you need the return value's format to match the [`search_for`](https://github.com/GovWizely/webservices/pull/174/files#diff-7ae1034585424914bae40382262f2e59R21). You want to take the return value from either method and pass it to the view.

A cleaner way to do this is to adjust the return value of `fetch_all` to match `search_for`, then adjust the CSV code to use this new format. Please do that instead.

spec/api/v2/market_research_library/search_spec.rb
calumhalcrow: When testing locally, I only got three results in the JSON:

screen shot 2015-07-08 at 1 42 58 pm

The application appeared to fetch all though (I could see the output of the ES query in the foreman output). Can you check this also?

calumhalcrow:
I rebased and force pushed FYI

app/models/concerns/searchable.rb
baraabourghli: :+1:

spec/api/v2/market_research_library/search_spec.rb
baraabourghli: Will check this.

baraabourghli:
All changes are done, also in my local I checked and it returns all results now.
Running codeclimate analysis.

PR #176, opened 2015-07-13
Envirotech

baraabourghli:


app/importers/envirotech/solution_data.rb
```diff
@@ -0,0 +1,61 @@
...
+    def current_page_empty?
+      @page ||= 1
+      @page_data = JSON.parse(Envirotech::Login.mechanize_agent.get(@resource + "?page=#{@page}").body)
+      @page += 1
+      @page_data.blank?
+    end
+
+    def current_page_data
+      @page_data
+    end
```
calumhalcrow: I feel a little uneasy about how the code works here in `current_page_empty?` and `current_page_data`. `current_page_empty?` asks a question, but it loads the next batch of documents as a side effect. Side effects in general are a sign of trouble.

I think the logic may be better like this:
```
def fetch_data
  page = 1
  result = []
  loop do
    page_data = JSON.parse("...?page=#{page}...")
    break if page_data.blank?
    result.concat(page_data)
    page += 1
  end
  result
end
```
With this code, there is not sharing of state between methods.

Also, since this code is repeated in each importer, you should introduce an Envirotech::BaseData class so that common functionality like this can be shared.

app/models/envirotech/mappable.rb
```diff
@@ -0,0 +1,72 @@
...
+            abstract_spanish:    { type: 'string', analyzer: 'spanish_analyzer' },
+
+            source_created_at:   { type: 'date', format: 'YYYY-MM-dd' },
+            source_updated_at:   { type: 'date', format: 'YYYY-MM-dd' },
+            source:              { type: 'string', analyzer: 'keyword' },
+            url:                 { type: 'string' },
+            source_id:           { type: 'string' },
+          },
+        },
+      }.freeze
```
calumhalcrow: At this stage, we don't have any fine-grained requirements about how each endpoint should behave. But it's worth keeping in mind that this isn't quite the same as other consolidated endpoints where several sources are made available via the same endpoint. What we have is several similar sources being made available via several similar endpoints. So it might happen that having the same mappings for each index doesn't work.

baraabourghli: Yeah I agree, it is worth keeping that in mind for future, but at this stage I only can see texts in several languages that need to be analyzed in the same way.

app/importers/envirotech/solution_data.rb
baraabourghli: >  Side effects in general are a sign of trouble.

Yep, need to improve this, will try your suggestion.


>since this code is repeated in each importer, you should introduce an Envirotech::BaseData class so that common functionality like this can be shared.

Yeah I have this in mind  :+1:

baraabourghli:
https://codeclimate.com/github/GovWizely/webservices/compare/envirotech#issues

CodeClimate GPA `-0.01`, but what can we do if it suggest that there are code duplicate because two links for the endpoints are almost similar? 
https://codeclimate.com/github/GovWizely/webservices/Envirotech::AnalysisLinkData?to_sha=8c6acf48

calumhalcrow:
@baraabourghli don't worry too much about the CC score. If it drops because of that, so be it.

PR #178, opened 2015-07-18
ita zip codes: sort by zip_code when relevance ties

carloslima:


spec/support/importer.rb
```diff
@@ -17,7 +17,7 @@
 shared_examples 'an importer which indexes the correct documents' do
   it 'indexes the correct documents' do
     expect(importer.model_class).to receive(:index) do |indexed_docs|
-      expect(indexed_docs).to eql(expected)
+      expect(indexed_docs).to match_array(expected)
```
carloslima: I don't think the order is relevant when checking if the correct documents got indexed, so I've changed this one.
Working around the order difference from the importer and the expected results seems like unnecessary noise.

Are my assumptions correct?

spec/spec_helper.rb
```diff
@@ -29,7 +29,6 @@
 
 RSpec.configure do |config|
   config.before(:suite) do
-    Webservices::Application.model_classes.each(&:recreate_index)
```
carloslima: This is unrelated to zip codes sorting but also a one liner, so I put it together :-P

From the commit message:

    Stop recreating indexes at the beginning of the test suite
    
    Tests already do this for the indexes they need, this one is redundant
    and causes running a single spec a lot slower since it needs to wait for
    all unrelated indexes to be recreated.

carloslima: I take back the one liner, but those were bugs, so I guess a fix is warranted regardless. :P

spec/support/importer.rb
calumhalcrow: Yeah, you're correct.

calumhalcrow:
What I find weird is that ES returns different scores when documents are the same except for the seemingly irrelevant "zip_code" field:

```
GET development:webservices:ita_zip_codes/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "multi_match": {
            "fields": [
              "post",
              "office_name",
              "zip_city"
            ],
            "operator": "and",
            "query": "guaynabo"
          }
        }
      ]
    }
  }
}
```

screen shot 2015-07-20 at 3 04 46 pm

It's a bummer because it makes the sort-by-zipcode changes ineffective (which means they'll probably be rejected by POs).

Any thoughts on this Carlos? We _might_ have to remove the sort by _score in this case :-(

PR #179, opened 2015-07-18
Stop recreating indexes at the beginning of the test suite

carloslima:
Tests already do this for the indexes they need, this one is redundant
and causes running a single spec a lot slower since it needs to wait for
all unrelated indexes to be recreated.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -29,7 +29,6 @@ Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
 
 RSpec.configure do |config|
   config.before(:suite) do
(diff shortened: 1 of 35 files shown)


carloslima:
I thought I'd make your life easier and put this as a separate PR :-)

spec/spec_helper.rb
```diff
@@ -29,7 +29,6 @@
 
 RSpec.configure do |config|
   config.before(:suite) do
-    Webservices::Application.model_classes.each(&:recreate_index)
```
carloslima: That's the change I wanted to make.

The rest are fixes for tariff rates specs which were testing queries against the wrong index.

PR #181, opened 2015-07-27
Tim/taxonomy

tmhammer:


calumhalcrow:
The code works as expected, but why do all the ID have stanford.edu in them? Is Stanford related to this endpoint in any way?

screen shot 2015-07-28 at 3 05 34 pm

If not, I suggest you find a more suitable ID value for each doc.

tmhammer:
@calumhalcrow  That prefix is automatically assigned to the id when you create a Class in Webprotege... not sure why.... we could remove it and keep the unique part of the id.  You can see them displayed under IRI when you click on a term:  http://52.4.82.207:8080/webprotege/#Edit:projectId=0aa09276-58a6-4350-a0bb-60eb2ab4be00

calumhalcrow:
@tmhammer it seems weird to mention stanford in this endpoint IMO. Why not ignore the given ID and generate one from all the document fields using `Utils.generate_id` ?

tmhammer:
@calumhalcrow  That sounds good to me.  I think the only reason I kept those ids in there originally was so there would be something unique to tie each JSON entry to its Class in Webprotege, in addition to just the name of the term.

PR #185, opened 2015-08-01
Carlos/last updated in results

carloslima:


spec/models/concerns/searchable_spec.rb
```diff
@@ -1,33 +1,55 @@
...
-          ['test_index_type']
+        self.model_classes = [MockModel]
+        def self.stored_metadata
+          {version: 123, time: 'a few minutes ago'}
         end
       end
 
-      # I'm doing this without the use of a class which includes Indexable
-      # in order to only use Searchable in this spec.
-      ES.client.indices.delete(index: index_name) if ES.client.indices.exists(index: index_name)
```
carloslima: @calumhalcrow Just want to draw attention to the fact that I've replaced this with actually using Indexable ;-)

carloslima:
@calumhalcrow 
All polished up and ready to go.
I squashed all the things that made sense to squash.
I believe the remaining ones make better sense as separate commits, as they are now. :-)

PR #186, opened 2015-08-01
Avoid recreating & importing envirotech index before each spec

carloslima:


carloslima:
It's easier to review if you use https://github.com/GovWizely/webservices/pull/186/files?w=1

carloslima:
Some failures still happening, I believe due to the date Mock. Will check later.

PR #187, opened 2015-08-05
Transliterate description field

tmhammer:


calumhalcrow:
@tmhammer I realize this is blocked and so maybe pending further changes, but I'm not seeing how transliterating would solve an encoding problem (which presumably is the cause of bad characters appearing in results). What's the underlying problem here? Can you give an example of a document which currently contains these bad chars (or point me to a story)?

calumhalcrow:
@tmhammer can I close this?

PR #194, opened 2015-08-12
Calum/envirotech spec speed improvements

calumhalcrow:


calumhalcrow:
@carloslima this solves the problem of recreating Envirotech indices on every example. It actually uses less rather than more mocks, by just adding the ability to pass a file path to an importer. By removing RSpec-specific code from the before blocks (in the case some mocking) I was able to convert them into `before(:all)` blocks. Take a look and see how much this speeds up your spec run.

I also made a few other changes, mostly just removing some redundant code. If you think it all looks ok, please go ahead and merge.

app/importers/envirotech/background_link_data.rb
```diff
@@ -21,7 +21,7 @@ def initialize(resource = ENDPOINT)
     end
 
     def import
-      data = fetch_data
+      data = @resource =~ URI.regexp ? fetch_data : JSON.parse(File.open(@resource).read)
```
carloslima: What I don't like about this approach is that this is essentially behavior that only happens during tests, while the fetch_data is only run in production.

I'd rather the "from_web" returned a resource file which could be passed in so the behavior would be the same.
2 steps, but the same. It would also be isolated and testable.

spec/importers/envirotech/base_data_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    allow(agent).to receive(:submit) do
+      field_with = double('value=' => nil)
+      login_form = double(field_with: field_with, buttons: [])
+      double(form: login_form)
+    end
+
+    allow(agent).to receive(:get).with('http://example.com?page=1') do
+      double(body: File.open(fixtures_file).read)
+    end
+    allow(agent).to receive(:get).with('http://example.com?page=2') do
```
carloslima: We should make this run for 3 pages at least if we want to spec the behavior of multiple page results.

carloslima:
I've made some comments on what I think could be improved here but those points are also valid for the current version.
This PR is already a big improvement and does make my spec runs considerably faster so I will just go ahead and merge it :-)

PR #200, opened 2015-08-17
Envirotech relational data

baraabourghli:


calumhalcrow:
Note to self: ensure CC score doesn't drop too much.

lib/tasks/envirotech.rake
```diff
@@ -0,0 +1,7 @@
+namespace :envirotech do
+  desc 'Update issue_solution_regulation.json file with latest data from toolkit scraper'
+  task update_local_relational_data: :environment do
+    $stdout.reopen("data/envirotech/issue_solution_regulation.json", "w")
+    puts JSON.pretty_generate(Envirotech::ToolkitScraper.new.all_issue_info)
```
calumhalcrow: This seems like a rather exotic way to [write to a file](http://alvinalexander.com/blog/post/ruby/how-write-text-to-file-ruby-example) :smiley:

spec/importers/envirotech/toolkit_data_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    context 'scraper fallback to local data' do
+      let(:all_issue_info) { }
+      let(:local_data) { { 'dummy' => 2 } }
+      it 'notify Airbrake when UI tool disappear' do
+        expect(Airbrake).to receive(:notify).with(Exceptions::EnvirotechToolkitNotFound.new)
+        importer.import
+      end
+    end
+  end
+end
```
calumhalcrow: Here's what I'm seeing here: `Envirotech::ToolkitData#import` is doing two things. 1. it kicks off all individual Envirotech imports and 2. it figures out if the scraped data is ok (and tells us about it if not). What your spec is testing is a) if `#import` chooses the correct relational data and b) if `Airbrake#notify` is called when it should be. Everything you're testing is related to point 2., and in order to do those tests, you have to mock everything related to 1. This creates a large amount of mocking code, which is a hint that something is not quite right here.

Why don't we break `#import` into two, one part handling the individual imports and the other fetching and checking the relational data? That way, you can write specs for the second method and avoid having to write any mocks at all! In fact, in my enthusiasm I actually started doing that [here](https://github.com/GovWizely/webservices/commit/816a4fdec0c0bb63516e0a0af9cf477bbea4ee20). Can you review my code, and then rework the spec so that it tests the new `fetch_relational_data` method?

In more general language what has happened here is a useful tool (mocking) has got in the way of good object-oriented principles (small, focused methods). It's a subtle trap to fall into, and you should look out for it. The sheer amount of mocking in the spec is the clue that exposes the problem.

calumhalcrow:
Ok here comes the rest of my obscene amount of feedback:

- How do I import? `rake ita:import[Envirotech::ToolkitData]` fails. `rake ita:import[Envirotech]` kicks off 7 separate Sidekiq jobs, but from Envirotech::ToolkitData I suspect that’s not how you wish the importer to work.
    I would like to maintain the `rake ita:import[Module]` syntax here, as it has a consistent meaning (which is: run all importers within the given module). That doesn’t mean it has to fire off independent Sidekiq jobs; if we must run sequentially then that’s fine. I just want the interface to be “run `rake ita:import[Envirotech]` in order to import all Envirotech data”. Can you make that happen?
    So right now I’m trying to figure out what this class which imports all envirotech data should be called and where it should live. Is it an importer? Yes. So it should live in app/importers/envirotech. What should it be called..? I’m not sure. *But*, we have logic which imports all data for a given module. This logic is buried in lib/ita.rake at present. So, here’s what I propose we do: move that logic out of the rake task and into a method in the Importers mixin called “import_all”. By default this will implement the current behavior. But we can overwrite the method on the Envirotech module and have it perform our synchronous import. Tell you what: I’ll take care of this tomorrow morning. It’s beyond the scope of the Envirotech work.

- I cannot see the relationships in the API results. They need to be there in order to build the widget.

- Talking about relationships, I’m trying to figure out exactly what they are. You have got `solution <— issue —> regulation` (where —> means has-many) but I don’t think that’s correct. If you go to the [UI tool](https://new.export.gov/envirotech/toolkit) and examine the “Groundwater Remediation” issue, based on how that example behaves I think the relationships are `issue —> regulation <—> solution` (the “Pump and Treat” solution shows up in both regulations, hence the many-to-many). Can you examine the data and confirm what the relationships are, then adjust the scraper and imports accordingly?

- The test coverage isn’t 100%.

- IMO Envirotech::Login should live in app/services. It doesn’t import anything, so having it in app/importers seems incorrect.

baraabourghli:
- You are write, i didn't expect this behavior of `Envirotech::ToolkitData`, will change it.

- True I will add it.

- I think it is a `many-to-many` relation, this mean either 1) i need to have a separate index for this relation, or 2) i need to add the IDs of the relations while importing, this will introduce the following complexity: 
I need to import the two tables, then i need to get the relation data and update the documents with list of mapped IDs.
I am not sure how elegant is solution (1) will be in the importer code, but defiantly it will be easier to query in the UI tool, so what do you think?

- I will change the code and add full coverage.

- Yep, will move that code to app/services.

calumhalcrow:
On the many-to-many relation discussion: we must define the relationships correctly, as per what we can deduct from the UI tool. If we need to add another index to store the many-to-many relationships, that's fine. If it makes the import more complex, that's also fine.

BTW, I suggest we don't expose the many-to-many info via a separate endpoint; we can get away with denormalizing it into arrays in the existing endpoint, like so:

```
issues:
[
  { id: 1,
    solution_ids: [1, 2] },
    id: 2,
    solution_ids: [2, 3] }
]
solutions:
[
  { id: 1,
    issue_ids: [1] },
  { id: 2,
     issue_ids: [1, 2] },
  { id: 3,
     issue_ids: [2] }
]
```

baraabourghli:
Yep @calumhalcrow, that is what i mean by solution (2), i think you made it clear that it is fine with the added complexity of the importer :+1:

calumhalcrow:
[This is the PR](https://github.com/GovWizely/webservices/pull/202) that introduces a "import_all_sources" method to modules containing imports, to be overwritten in `Envirotech`. Probably best that I review the code myself in the morning before assigning it to anyone (my brain's functionality is fading right now).

baraabourghli:
I have done some improvements, I also reviewed what you did in https://github.com/GovWizely/webservices/pull/202 and it is straightforward. I will continue improving the code once you are done with it : )

calumhalcrow:
One more thing: you need to add VersionableResource to all the importers.

calumhalcrow:
@baraabourghli #202 has been merged. Let me know when you've acted on all the feedback I've given you and I'll review again.

baraabourghli:
@calumhalcrow I guess this should be fully working now, please take note that I introduced partial-update for ES index in order for us to import + denormalize the many-to-many relations using the ToolkitScraper, hopefully it is okay to do importing that way until we have full access to the APIs.

calumhalcrow:
Looks much better now Baraa. There are still a couple of things, then we can consider this done:

- The source code of the admin app has many-to-many relationships between issues, regulations and solutions. We don’t have that at present. Can you change it so that we have that also (I think it will be necessary in order to make the widget work exactly as how the current “toolkit" works).
- Check out this URL: https://api.govwizely.com/envirotech/foo/search?api_key=gMd2zNQE_r9Wwpj-l7NjGY9b&sources=bar
    There’s a couple of problems with it: 1) foo isn’t a valid Envirotech endpoint. 2) sources shouldn’t be an allowed parameter.

Also note that I pushed a couple of minor changes.

calumhalcrow:
I found one other thing that needs to be changed: the `envirotech:update_local_relational_data` rake task.

Right now it relies on ES being populated with issues. If there are no issues in ES it will set the JSON file to an empty object. Instead it needs to just grab all information from the UI tool regardless of what's in ES (it shouldn't query ES at all in fact). Right now, it saves the issue's ID as stored in ES to the file, but how do we know that those IDs will be the same in all dev envs? We should match issues grabbed from the admin tool to ones scraped from the UI tool by name after the scraping has been done.

calumhalcrow:
@baraabourghli I have listed outstanding issues on [the story](https://www.pivotaltracker.com/n/projects/1085558/stories/98487958) so that we don't forget about them :smiley:

calumhalcrow:
@baraabourghli this is how I would make the necessary changes:

1. Update Envirotech::ToolkitScraper to reflect our new-found understanding of the data model. While doing this, remove the need to look up issues from ES. We now need to view each issue, *then* each regulation for that issue. We’d end up with Envirotech::ToolkitData creating a data structure like this:
```
“issue name 1”: [
  “regulation name 1”: [
    “solution name 1”, “solution name 2”, “solution name 3”,
  ],
  “regulation name 2”: [
    “solution name 2”, “solution name 4”, “solution name 5”,
  ],
],
“issue name 2”: [
  “regulation name 3”: […]
],
...
```
This would get save to the JSON file etc.

2. Create a service which has the following methods:
```
Envirotech::Relationships.new.regulations_for_issue(“issue name 1”)
  # returns [“regulation name 1”,  “regulation name 2”]
Envirotech::Relationships.new.solutions_for_issue(“issue name 1”)
  # returns [“solution name 1”, “solution name 2”, “solution name 3”, “solution name 4”, “solution name 5”]
Envirotech::Relationships.new.solutions_for_regulation
Envirotech::Relationships.new.issues_for_regulation
Envirotech::Relationships.new.issues_for_solution
Envirotech::Relationships.new.regulations_for_solution
```
They should all return arrays of names, and use only the scraped data (either JSON file or actual network communication with the toolkit). There should be no communication with ES.

3. Use the service in the importers to match issue, regulation and solution documents with each other. When done, each type of document will have two “???_ids” fields, where ??? = each of the other two document types. The fields should be arrays of integers.

For example, there would be code like the following somewhere in the solutions importer:
```
solution[‘issue_ids’] = Envirotech::Relationships.new.issues_for_solution(solution[‘name’])
solution[‘regulation_ids’] = Envirotech::Relationships.new.regulations_for_solution(solution[‘name’])
```

That should do it.

baraabourghli:
Okay understood, then what I just did for mapping solutions and regulations is similar to what you have in mind, I need to adjust the scrapper to get data without ES query and create the new `Relationships` class to manage the mapping.

baraabourghli:
This branch as it is now only missing the prevention of calling non existing `sources` and passing `sources`in query param. I think you mean that we need to convert it to separate APIs not consolidated correct?

calumhalcrow:
@baraabourghli thanks for these new changes, I'm going to review them now.

About the routes: I don't think we need to do too much work to fix this. Let me take a look over the routes/controller code first before I comment further...

app/importers/envirotech/relational_data.rb
```diff
@@ -0,0 +1,111 @@
...
+    def import
+      return unless  RelationalData.relations.present?
+      process_solution_relations
+      process_regulation_relations
+    end
+
+    private
+
+    def process_regulation_relations
+      regulations = Envirotech::Consolidated.search_for(sources: 'regulations', size: 100)
```
calumhalcrow: Throughout our discussion I've been surprised that you start the matching of related documents by fetching a list of documents from ES. I was expecting that you'd start by iterating over the fetched JSON, like in this pseudo-code:
```
json = Envirotech::ToolkitData.fetch_relational_data
json.keys.each do |issue_name|
  issue = ES.fetch(issue with name "issue_name")
  json[issue_name].each do |regulation_name|
    regulation = ES.fetch(regulation with name "regulation_name")
    issue.regulation_ids << regulation.id
    regulation.issue_ids << issue.id
    json[issue_name][regulation_name].each do |solution_name|
      solution = ES.fetch(solution with name "solution_name")
      regulation.solution_ids << solution.id
      solution.regulation_ids << regulation.id
      solution.save
    end
    regulation.save
  end
  issue.save
end
```
Doing this avoids the problem when there are more than 100 issues or regulations. Do you think iterating over the JSON is better, or is there a problem doing so that I'm not seeing?

calumhalcrow: Not really related, but I'll put it here anyway: I ran a full import locally. One of the issues didn't have any associated regulations, but the toolkit shows that it does:
screen shot 2015-09-15 at 11 34 09 am

calumhalcrow:
@baraabourghli one more thing: can you rebase and rubocop -a? After rebasing specs will fail; you'll have to fix that.

app/importers/envirotech/relational_data.rb
baraabourghli: I am looking into your first suggestion, in the meantime, I fixed an issue with HTML escape characters, relational importing should be working correctly for now.
Also there was some data changes from the source, I updated the local data.

PR #202, opened 2015-08-19
Calum/import all

calumhalcrow:


calumhalcrow:
ok @baraabourghli, go crazy...

app/importers/concerns/can_import_all_sources.rb
```diff
@@ -0,0 +1,23 @@
...
+      Dir[module_importer_files].each { |f| require f }
+
+      constants
+        .map { |constant| const_get(constant) }
+        .select { |klass| klass.include?(Importable) && !klass.disabled? }
+    end
+
+    def import_all_sources
+      importers.each do |i|
+        next if i.disabled?
```
baraabourghli: Isn't this check duplicate of `!klass.disabled?` in `importers`?

app/importers/concerns/importable.rb
```diff
@@ -1,10 +1,16 @@
...
-  def self.included(base)
-    base.class_eval do
-      base.send(:prepend, ParentInstanceMethods)
+  included do
+    send(:prepend, ParentInstanceMethods)
+
+    class << self
+      attr_accessor :disabled
+      def disabled?
+        !!disabled
```
baraabourghli: What do you think of using `.present?` instead? i am curious to know why not :)

calumhalcrow: You mean you don't like the "bang bang you're boolean" operator?!? :smile: 

In all practical use cases `disabled` will be either `true` or `nil` (it might get set to `false`, but other than that nothing else I'd hope). I could add more code that defaults `disabled` to false, and also validates its value, but I don't see any point in doing that. If you put `self.disabled = 'donkey'` in your importer then you should expect problems :smiley:

`present?` is a Rails thing, and is aimed at strings that might be empty or contain only whitespace. Since I'm dealing with only `nil` and `true`, it seems a bit too sophisticated for our needs. The double-bang is there to just convert `nil` to `false`, so that `disabled?` returns a boolean. And it's pure Ruby code, which probably makes it faster although I didn't break out the profiler so I don't know that for sure :wink:

app/importers/concerns/can_import_all_sources.rb
calumhalcrow: Yes, it is. Good catch.

app/importers/concerns/importable.rb
baraabourghli: Awesome, I just needed to know why so I can decide what to do next time I encounter a similar situation. and i like double-bang :+1:

baraabourghli:
:+1:

PR #213, opened 2015-09-07
Laiq/make remaining importers versionable

laiqjafri:
A few importers made versionable

### DIFF
--- a/spec/support/importer.rb
+++ b/spec/support/importer.rb
@@ -23,3 +23,14 @@ shared_examples 'an importer which indexes the correct documents' do
     importer.import
   end
 end
+
+shared_examples 'a versionable resource' do
+  it 'updates version properly' do
+    expect(importer.stored_version).not_to eq importer.available_version
+    importer.import
+
+    # When resource is unchanged, stored_version should be equal available_version
+    new_importer = importer.class.new(resource)
+    expect(new_importer.stored_version).to eq new_importer.available_version
+  end
+end
--- a/spec/support/importer.rb
+++ b/spec/support/importer.rb
@@ -8,12 +8,6 @@ shared_examples 'an importer which cannot purge old documents' do
   it { is_expected.to be_falsey }
 end
 
-shared_examples 'an importer which versions resources' do
-  it 'has a non-abstract #available_version method' do
-    expect(described_class.new.method(:available_version).owner).to_not be(Importable)
-  end
-end
-
 shared_examples 'an importer which indexes the correct documents' do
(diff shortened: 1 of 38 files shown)


spec/support/importer.rb
```diff
@@ -23,3 +23,16 @@
...
+
+  it 'updates version properly' do
+    expect(importer.stored_version).not_to eq importer.available_version
+    importer.import
+
+    # When resource is unchanged, stored_version should be equal available_version
+    new_importer = importer.class.new(resource)
+    expect(new_importer.stored_version).to eq new_importer.available_version
+  end
+end
```
calumhalcrow: This looks great so far @laiqjafri :+1: Just a couple of things:

- We had trouble in the past with recreating indices on every example (i.e. in `before` blocks). The operation can greatly slow down the spec run if it happens too often, and it's better if we can restrict it to `before(:all)` blocks instead. In order to help you with this pushed a couple of commits: [one](https://github.com/GovWizely/webservices/commit/cf784ff0988d63387ebd88055c8a7d354b88954e) to add `model_class` as a class method to anything Importable, [an addition to this shared example](https://github.com/GovWizely/webservices/commit/eaa2dcba8cc6d08d434e77c667917d661568b09a) and a [modification to the trade articles importer spec](https://github.com/GovWizely/webservices/commit/a0649c05a9ab23fca5fc0a889045b342f431fcee) showing what I mean. Can you ensure that you only add recreate_index calls in `before(:all)` blocks in other specs?
- There's an existing shared example called 'an importer which versions resources' which does something similar to your new shared example (albeit in a much more cryptic way). Can you combine both into one so that we don't have two of them?

calumhalcrow: Also, can you rebase this branch onto master to pick up [these changes](https://github.com/GovWizely/webservices/pull/207)? They are somewhat related to the work in this PR.

spec/support/importer.rb
```diff
@@ -23,3 +17,20 @@
     importer.import
   end
 end
+
+shared_examples 'a versionable resource' do
+  before { described_class.name.gsub(/Data$/, '').constantize.recreate_index }
```
calumhalcrow: You should be able to just do `described_class.model_class.recreate_index`.

calumhalcrow: In fact, I'll change it myself... the rest of the PR looks fine, but I'll do some thorough testing before merging.

laiqjafri: @calumhalcrow cool!!

PR #217, opened 2015-09-22
Separate fields for mapped and original Industries

tmhammer:


tmhammer:
https://www.pivotaltracker.com/story/show/101435114

tmhammer:
I think this should be ready now @calumhalcrow.  I noticed the "industries" search was only searching the original industries so I updated it to search the ITAs as well.

PR #220, opened 2015-10-01
Tim/urls csl

tmhammer:


calumhalcrow:
Things that need addressing:
- [ ] Is there anything special we need to do regarding copying the URL mapping index to staging/production?
- [ ] can we use the Rails cache instead of an instance variable? (url_lookups method)
- [ ] IMO it would be cleaner to have a service to look up Bitly URLs rather than adding the get_bitly_url method to Importable.
- [ ] @jamesisaacs Where are we on merging calum/fully-decoupled-staging? Because I'd very much like not to add a environment variable (BITLY_ACCESS_TOKEN) to the project at this stage.
- [ ] How can I get the BITLY_ACCESS_TOKEN value, so that I can add it to the staging stack?
- [ ] Ideally I'd like to avoid adding "complex methods" (as reported by CodeClimate).
- [ ] (Nice to have) We could keep a record of the last time we called the Bitly API and only sleep if we needed to.
- [ ]  bitly_api_request - put this in config? Bitly gem?
- [ ] `rubocop -a
  why the failing specs locally?

tmhammer:
@calumhalcrow  I think I've addressed almost everything... The Bitly gem looked promising but it doesn't look like it has anything for saving metadata along with links.  I also left out the "recording request times and sleeping when needed" for now in the interest of time, the CSL only has 17 URLs so it should be fine.  Although this is something I still want to do, and can maybe add it in when we merge URL tracking for another endpoint.  Also the rake commands for managing the URL index:  `bundle exec rake urls:recreate_index` and `bundle exec rake urls:purge_old`

app/importers/concerns/importable.rb
```diff
@@ -1,3 +1,5 @@
+require 'open-uri'
+
```
calumhalcrow: Why is it necessary to add this to Importable?

calumhalcrow:
I'll put this on staging and take care of the Chef recipe changes needed in order to get the token into the config.

What we need though is for the specs to pass without requiring the token to be set. We cannot have developers requiring the token in order to do unrelated development work. I'll take a closer look at the code and figure out how to make this happen...

tmhammer:
Good point @calumhalcrow, I think this should work.

PR #226, opened 2015-10-08
return not found if :sources is invalid

viclim:


calumhalcrow:
@viclim I was thinking that the solution to this would be via a constraint in `config/routes.rb`. Something like:

```
namespace :envirotech do
  get ':sources/search', to: 'consolidated#search',
    constraints: lambda { |request| %w(solution issue etc).include?(request.params[:sources]) }  # pseudo-code
end
```

viclim:
@calumhalcrow 
Is it possible to include multiple sources in the API? e.g. `/envirotech/solutions,issues/search`
If it is not then a route constraints will be a perfect fit.

calumhalcrow:
@viclim we don't want to support multiple sources via the URL, so yeah you can use constraints here.

PR #227, opened 2015-10-08
Tim/new mrl urls

tmhammer:


app/importers/market_research_data.rb
```diff
@@ -3,9 +3,9 @@
 
 class MarketResearchData
   include Importable
-  include VersionableResource
```
calumhalcrow: You shouldn't need to remove this, and it's better if you can keep it in place TBH. Since you have two sources, you need to override `available_version` so that it returns a meaningful value. Take a look at [SBA Trade Events](https://github.com/GovWizely/webservices/blob/master/app/importers/trade_event/sba_data.rb#L24), which is similar to what you want to do.

tmhammer: @calumhalcrow, if the source is no longer going to be updating, won't doing the version check make it so the importer never runs?  We still need it to catch Industry Mapper updates.

calumhalcrow: Ok, good point. In fact, I need to think about how VersionableResource is going to work with Industry Mapper in the mix... feel free to ignore my comment.

PR #229, opened 2015-10-09
Bug fix market research library industry filter

viclim:


viclim:
@calumhalcrow I think this is how the escaped comma can be implemented to fix the issue.

app/queries/v2/market_research_query.rb
```diff
@@ -3,7 +3,9 @@ def initialize(options = {})
     super
     @sort = @q ? nil : 'title.keyword'
 
-    @industries = options[:industries].split(',').map(&:strip) if options[:industries].present?
+    @industries = options[:industries]
+      .split(/(?<!\\),/)
```
calumhalcrow: Nice use of a [negative lookbehind assertion](http://docs.ruby-lang.org/en/2.2.0/Regexp.html#class-Regexp-label-Anchors), it does exactly the right thing here :smile:

PR #231, opened 2015-10-12
Refactor es model custom analyzer definition

viclim:
Refactor by creating an `analyze_by` macro in indexable concern, and a `Analyzers` module to store definitions.


calumhalcrow:
The only reason I haven't merged this is because when I rebased other [open PRs](https://github.com/GovWizely/webservices/pulls) onto it their specs started failing. I want to find out why that is before merging...

PR #235, opened 2015-10-21
Administrator can automagically generate a search API from an uploaded CSV file

loren:
This is an early, untested prototype of the CSV->API builder. I'd like to get it up and running in a staging environment for Stuart and others to try out. In order to do this, I think all you need to is set the new `is_admin` flag to true for whatever user account Stuart will use to log in. If there's just one user, this will do the trick:

```
[5] pry(main)> User.all.first.update_attribute(:is_admin, true)
```

Comments/questions welcome on the approach, etc, too.


calumhalcrow:
@loren this is now on staging. I went through the list of users on staging and upgraded to admin everyone I thought necessary. I successfully created an endpoint, so it all appears to be working from my point of view.

loren:
Wow awesome! Thanks.

README.md
```diff
@@ -56,9 +56,16 @@ Generate an admin user:
...
     bundle exec rake db:devseed    
 
-Fire up a server and import some data.
+Fire up a server:
 
     foreman start -f Procfile.dev
+    
+Or, if you already have Elasticsearch and Redis running because the web and sidekiq actions timeout in `Procfile.dev`, you can start foreman like this:
+
+    foreman start -f Procfile_no_es_redis.dev
```
calumhalcrow: Since this PR causes the Rails app to fetch info from ES during initialization, I suspect devs will have to start ES first before running foreman, i.e. we'll likely end up renaming Procfile_no_es_redis.dev to Procfile.dev.

loren: I ran into a couple of different/other issues with foreman's interaction with elasticsearch and redis. The only way I could get things to work was to have redis and ES already running. But I'm open to moving that initializer stuff somewhere else. The first line just ensures the DataSource index is created in ES regardless of where it's run. The rest of that file just writes and executes the Ruby code that sets up the persistence models.

loren:
I moved `api_models.rb` over to lib so all that class definition stuff isn't happening during the initialization phase when the elasticsearch connection is still getting set up.

calumhalcrow:
Hi there. I'll summarize my feedback here, so that we can keep track of it etc.

- [ ] Need to rebase this branch onto master to pick up recent commits.

README.md
calumhalcrow: I don't think there's any getting round the fact that during initialization Rails has to query ES in order to know which model classes to create. I don't see any problem with that being the case. I'd rather we just have one Procfile for development if possible, so I suggest moving Procfile_no_es_redis.dev to Procfile.dev and updating the README so that it tells devs to have ES and Redis running before they run foreman.

app/assets/stylesheets/application.css.scss
```diff
@@ -779,5 +779,14 @@ font-size:6px;
...
 
+textarea#data_source_dictionary {
+  width: 100%;
+}
+
+.data_source .field input
+{
+  width: 50%;
+}
+
```
calumhalcrow: Better to put this in [customization.css.scss](https://github.com/GovWizely/webservices/blob/master/app/assets/stylesheets/customizations.css.scss). The contents of this file are copied verbatim from developerportal; any webservices-specific modifications are in customizations.css.scss. This way, if developerportal CSS is updated we can just copy it here without worrying about deleting modifications. Not a great system but that's what we have.

app/models/user.rb
```diff
@@ -124,6 +124,8 @@ def email_was
   attribute :failed_attempts, Integer, default: 0, mapping: { type: 'integer' }
   attribute :unlock_token, String
 
+  attribute :is_admin, Boolean, default: false
```
calumhalcrow: Just a nit, but IIRC the `is_` prefix isn't very Rubyish. Maybe just `admin`?

spec/api/v1/envirotech/consolidated_spec.rb
```diff
@@ -236,7 +236,7 @@
     let(:params) { { size: 100 } }
 
     context 'when :sources are invalid' do
-      it { expect { get '/v1/envirotech/invalid_source/search', params }.to raise_error ActionController::RoutingError }
+      xit { expect { get '/v1/envirotech/invalid_source/search', params }.to raise_error ActionController::RoutingError }
```
calumhalcrow: Maybe just delete this example, since v1 has already been deprecated?

spec/features/data_source_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(page).to have_text("Some other human readable name")
+
+      click_link('Back')
+      expect(page).to have_text("Some other human readable name")
+
+      click_link("Some other human readable name")
+      click_button("Delete")
+      expect(page).to have_text("Dataset was successfully destroyed")
+
+    end
```
calumhalcrow: Nice :smiley:

loren:
I incorporated all your feedback (thank you!) and fixed an unrelated deprecation warning that was bugging me. I optimistically removed the `do not merge` tag.

calumhalcrow:
@loren I pushed a commit to this branch which applies the rubocop changes. I'm thinking of creating a git commit hook which ensures that rubocop is run before creating a commit. That way we'll all always incorporate such changes into our commits.

calumhalcrow:
@loren regarding merging to master: we tend not to merge such large features until the requester accepts the story in PT. Can you get Erik/Stuart to do that? When done I'll merge.

[We even documented a [story lifecycle](https://github.com/GovWizely/webservices/wiki/Story-Lifetime--and-Deployment-Process) process at one point!]

loren:
Got it. I should have asked if you guys had any guidelines or anything like that.  Want me to squash the endpointme commits into one big one and force push to the origin? 

Also, are you OK with adding this block of code https://github.com/GovWizely/webservices/blob/ed444986d28bdbbd5b1e5a932eb07c3975e9bcfa/config/environments/development.rb#L35-L42 to your `config/environments/production.rb` file? We'll need it for acceptance testing (and of course production deployment when we get there). If you have some other better way you want to handle this, I'm open to anything.

calumhalcrow:
@loren given that the commit messages are (mostly) all the same, I'd go with a squash.

Regarding the block of code: you'll have to add that to our Chef cookbook, [specifically here](https://github.com/GovWizely/webservices-cookbook/blob/master/templates/default/environments_application.rb.erb). I don't have any objection.

loren:
I squashed the commits (except yours @calumhalcrow) and rebased off master. I also added that block to the Chef cookbook and created a pull request (https://github.com/GovWizely/webservices-cookbook/pull/5).

loren:
The staging environment exposed a problem that didn't occur on dev or staging. I'm going to have to change how I go about dynamically defining the constants around the api models.

The problem can occur when you search on an API that you just created. If you create a new api called, say, de_minimis_currencies, the webservices code in that Rails process dynamically writes code to define a model called DeMinimisCurrency, and then evaluates that code. The next time a request comes in to that Rails process, it knows about DeMinimisCurrency and can process search requests for it. If you spin up another rails process (for Sidekiq, or the Rails console, or via Passenger), it will initialize DeMinimisCurrency before any requests are fielded. But, if you have multiple Passengers running (or multiple web machines running Rails) and you create an API, only that single Rails process will know about the DeMinimisCurrency. The other ones won't, and if they happen to service any subsequent search requests for the DeMinimisCurrency API, the request will fail. If you restart the other Rails procs, everything will work. The problem is even harder to track down if you happen to rename a field in the data dictionary. Search requests that don't reference that field will work on all Passenger instances, but requests that refer to the new field name will only work on one Passenger instance.

I think the thing to do here is to pessimistically unload and load--- on every search request---  the API model classes associated with data sources. I don't think it will end up being particularly slow, and we can always evaluate a more optimistic pub-sub strategy later. If anyone has suggestions on how to handle this better in the meanwhile, let me know.

I will make the necessary changes and let you know when it's ready.

calumhalcrow:
@loren feel free to squash my commit also; all it does is apply auto-generated cody tidying provided by rubocop.

Amen for staging! The dynamic creation of a model class for each endpoint is on the one hand consistent with webservices' existing approach, but on the other hand problematic in this situation. The only way I can think to avoid this situation is to not dynamically create classes at all, but that would lead to a completely different approach (you'll be starting again). So I don't have much more that what you've already figured out for yourself.

loren:
@calumhalcrow I squashed the rubocop commit. And yes, amen for staging. It saved a lot of production-grade headache.

I came up with an approach to the problem that is pretty lightweight and I actually like better than what I had there before. It's hard to see the difference as everything is squashed together but I introduced an instance method in `DataSource` called `with_api_model` that sets up the Ruby class based on the latest data dictionary for that data source. It gets called when indexing data and it gets called when searching.

calumhalcrow:
@loren I see what you did there :-) looks good.

calumhalcrow:
@loren I've just noticed something that should be adjusted before we merge this. The results format should be consistent with other endpoints. See the following:

screen shot 2015-11-10 at 5 52 06 pm

loren:
Do you mean having the total/offset/results fields as top level elements? I am doing that as part of https://www.pivotaltracker.com/story/show/107724824 <https://www.pivotaltracker.com/story/show/107724824> but could just merge it into this story.

calumhalcrow:
@loren if you're doing that as part of another story then that's fine, I didn't realize that that was the case.

calumhalcrow:
It appears that there's nothing stopping me from merging this to master. Shall I do that @loren ?

loren:
No, hang on and I will have something for you tomorrow.

PR #238, opened 2015-10-25
Envirotech import refactoring

baraabourghli:


baraabourghli:
this should be more straightforward to maintain. what do you say @calumhalcrow ?

calumhalcrow:
@baraabourghli yeah, this is the better approach. Thanks for implementing it.

Regarding `Array#uniq_push!`: I believe there's a way in which you can achieve the same effect but without monkey patching Array (it's better to avoid monkey patching core classes if possible). You can use Array unions (single pipe `|`):

`issue[:regulation_ids] |= [regulation[:source_id]]`

Can you back out the monkey patch and use this approach instead?

baraabourghli:
Okay, i agree :+1:

baraabourghli:
BTW @calumhalcrow it is done, i forgot to mention it : P

calumhalcrow:
Oh ok, thanks for pointing that out :-) will review shortly...

PR #239, opened 2015-10-26
Remove tag

tmhammer:


Gemfile
```diff
@@ -29,7 +29,7 @@ gem 'mechanize'
 gem 'sass'
 gem 'jquery-rails'
 
-gem 'industry_mapping_client', git: 'git://github.com/GovWizely/industry_mapping_client.git', tag: 'v0.1.1'
+gem 'industry_mapping_client', git: 'git://github.com/GovWizely/industry_mapping_client.git'
```
calumhalcrow: Any reason why you updated Gemfile but not Gemfile.lock @tmhammer @jamesisaacs ?

jamesisaacs: @calumhalcrow Just an oversight, but when adding Gemfile.lock I found my recently upgraded version of bundler appended the following version information to the file:

```+
BUNDLED WITH
   1.10.6```

As a workaround, I removed the lines, downgraded bundler to version 1.5.3 -- the version we're using on OpsWorks (the aws default), tested, and committed Gemfile.lock. 

I found some heated discussion of the 'BUNDLED WITH ...' issue online. I'll ping you about it later.

calumhalcrow: I read over [this discussion](https://github.com/bundler/bundler/issues/3697), and also this [bundler release notice](http://bundler.io/blog/2015/06/24/version-1-10-released.html#bundled-with). IMO this best thing to do is for all of us to upgrade to bundler 1.10.x. 1.5.3 is almost two years old; I don't know why it is OpsWorks' default. And on that point: I recently had to upgrade an OpsWorks stack from 1.5.3 to 1.10.6 in order to make the stack's deployment work with modern bundler syntax.

PR #240, opened 2015-10-28
Envirotech SolutionProviders sort

calumhalcrow:


app/importers/envirotech/analysis_link_data.rb
```diff
@@ -42,7 +42,7 @@ def process_article_info(article_hash)
 
       article[:source] = model_class.source[:code]
 
-      article[:id] = Utils.generate_id(article, %i(source_id source))
+      article[:id] = article[:source_id]
```
baraabourghli: Good to know that it is okay to use the relation id for ES id, i think we can now improve the `update` method in `Indexable`! will take a look at this later.

baraabourghli:
:+1:

app/importers/envirotech/analysis_link_data.rb
calumhalcrow: Yeah, since each source has it's own index, we can safely use the "source_id" as the document id. It makes `get`ting a document much easier over in the ProviderSolutions importer, which I've changed in this PR.

calumhalcrow:
@baraabourghli I've assigned this PR to you since you're familiar with the code.

baraabourghli:
Yeah, i am done with the review, it looks good to me.

PR #247, opened 2015-11-09
New source endpoint for USTDA Trade Events

tmhammer:


tmhammer:
https://www.pivotaltracker.com/story/show/98225300

calumhalcrow:
Some feedback:

- [ ] TradeEvents already has start_time and end_time fields (see SBA). You should use those field names rather than "event_time", and make sure they are in 24-hour format:
screen shot 2015-11-10 at 6 40 49 pm

calumhalcrow:
@tmhammer your changes look good. I have pushed a commit that should deal with the CC issues, but for some reason it reduces coverage. I'll fix that tomorrow my time.

jamesisaacs:
Waiting to hear back from USTDA why there are events from 2014.

PR #251, opened 2015-11-17
[#108246040] Blank screen on Endpoint Me!

loren:
- Added restrictions on the data source's api field
- rubocop'd api_models_controller_spec

### DIFF
--- a/spec/models/data_source_spec.rb
+++ b/spec/models/data_source_spec.rb
@@ -7,6 +7,14 @@ describe DataSource do
       is_expected.to validate_presence_of(:api)
       is_expected.to validate_presence_of(:data)
     end
+
+    ['foo bar', 'de-minimis', "loren's", 'Currencies'].each do |bad_value|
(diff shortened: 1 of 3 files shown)


spec/models/data_source_spec.rb
```diff
@@ -7,6 +7,14 @@
       is_expected.to validate_presence_of(:api)
       is_expected.to validate_presence_of(:data)
     end
+
+    ['foo bar', 'de-minimis', "loren's", 'Currencies'].each do |bad_value|
+      it { should_not allow_value(bad_value).for(:api) }
```
calumhalcrow: I suggest you use RSpec's newer `is_expected.to` syntax for one-liners. Even though "should" one-liners are not deprecated in RSpec 3 (as opposed to the should method money patch which is), using `is_expected.to` will make your changes consistent with the rest of the codebase.

loren:
I made the rspec change. Please have another look.

PR #258, opened 2015-12-02
Ita_Industries from Mapper for all Trade Leads endpoints

tmhammer:


tmhammer:
https://www.pivotaltracker.com/story/show/105665602

calumhalcrow:
@tmhammer is it possible to deal with the codeclimate issues?

calumhalcrow:
@tmhammer also: rubocop -a and rebase onto master please :-)

calumhalcrow:
One thought @tmhammer: since all trade lead importers now pull data from IM, should you remove `include VersionableResource` from them all? If it remains, if the taxonomy changes but the source doesn't, the endpoint will not be updated on import.

tmhammer:
Thanks for pointing that out @calumhalcrow, I ended up tweaking the Versionable logic slightly so we can make an importer still import even if the source doesn't change.  Which means we can keep using the versionable metadata on the endpoints.

calumhalcrow:
ok looks good.

PR #269, opened 2015-12-15
Tim/naics lookups

tmhammer:


Gemfile.lock
```diff
@@ -6,7 +6,7 @@ GIT
 
 GIT
   remote: git://github.com/GovWizely/industry_mapping_client.git
-  revision: 7d23ed06f9815081d8de30b45a564506a22c3d87
+  revision: 2c2b4bd4300f0b10b1e32bcf864c9e587b78e354
```
calumhalcrow: This is why I added version numbering to IMC :wink:

tmhammer:
@calumhalcrow  These commits are already on staging and should be ready to merge to master, pending acceptance of this story:  https://www.pivotaltracker.com/story/show/109756984.  I will also close the other PR since those changes are included here.

app/importers/concerns/importable.rb
```diff
@@ -90,6 +90,10 @@ def normalize_industry(industry)
     end
   end
 
+  def get_mapper_terms_from_array(array)
+    array.map { |i| normalize_industry(i) }.compact.flatten.uniq
+  end
```
calumhalcrow: I'm a huge fan of this. A better solution would be to add this functionality to the gem, i.e. you could pass an array of terms to IndustryMappingClient#map_industry and it would do the rest (including managing caching). It's possible to look up several terms at once via the IM API, and taking advantage of that would lead to the most efficient solution.

However, rather than ask you to make such changes, I've deferred them to [another story](https://www.pivotaltracker.com/story/show/110477482). It would be too much of a detour to ask you to do them as part of this PR.

app/importers/concerns/importable.rb
```diff
@@ -112,6 +116,12 @@ def parse_city_from_address(event_hash)
     end.compact.first
   end
 
+  def process_industries(entry)
+    entry[:industry] = entry[:industry] + ': ' + @naics_mapper.lookup_naics_code(entry[:industry]) if entry[:industry]
```
calumhalcrow: You cannot really do this. If you include this concern in a class which doesn't set `@naics_mapper` then calling this method will cause an error. Better to do:

```
class A
  include Importable
  attr_accessor :naics_mapper
  def initialize
    self.naics_mapper = NaicsMapper.new
  end
end

module Importable
  extend ActiveSupport::Concern
  def process_industries(entry)
     fail 'must implement naics_mapper' unless self.class.method_defined?(:naics_mapper)
     # original implementation
  end
end
```

app/importers/concerns/versionable_resource.rb
```diff
@@ -18,6 +18,7 @@ def import
         update_metadata(available_version)
         Rails.logger.info "#{self.class.name}: resource updated, new data indexed."
       else
+        super if self.class.const_defined?(:CONTAINS_MAPPER_LOOKUPS) # Run the importer anyways to catch Mapper updates
```
calumhalcrow: Instead of adding this trailing conditional, which confuses the if/else block somewhat, wouldn't it be better to adjust the first if? i.e.:
```
if resource_changed? || self.class.const_defined?(:CONTAINS_MAPPER_LOOKUPS)
```
You can interpret this to mean "do the import if the source document has changed or we need to do mapper lookups (which we cannot know have changed or not)". It's more representative of our thinking here therefore communicates the intent more clearly.

lib/naics_mapper.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize
+    options = {
+      user_provided_headers: CSV_HEADERS,
+      headers_in_file:       false,
+    }
+    @lookup_hash = SmarterCSV.process("#{Rails.root}/data/2012_to_2007_NAICS.csv", options)
+  end
+
+  def lookup_naics_code(code)
+    Rails.cache.fetch("#{code}", expires_in: 60.minutes) do
```
calumhalcrow: You're creating one instance of the class for use over the lifetime of the importer, and it gets it's data from a file on the local filesystem, which is read one time upon instantiation and then stored in memory...so I don't see the need to use the Rails cache here.

lib/naics_mapper.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def search_for_2012_code(code)
+    @lookup_hash.find { |entry| entry[:code_2012] == code.to_i }
+  end
+
+  def search_for_2007_code(code)
+    @lookup_hash.find { |entry| entry[:code_2007] == code.to_i }
+  end
```
calumhalcrow: Seems like one method, and passing the year as an argument, would lead to less code repetition? [Keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments) would be good.

spec/api/v1/australian_trade_leads_spec.rb
```diff
@@ -3,8 +3,10 @@
 describe 'Australian Trade Leads API V1', type: :request do
   before(:all) do
     TradeLead::Australia.recreate_index
-    TradeLead::AustraliaData.new(
-      "#{Rails.root}/spec/fixtures/trade_leads/australia/trade_leads.csv").import
+    VCR.use_cassette('importers/trade_leads/australia.yml', record: :once) do
+      TradeLead::AustraliaData.new(
+        "#{Rails.root}/spec/fixtures/trade_leads/australia/trade_leads.csv").import
+    end
```
calumhalcrow: You might find it cleaner in the long-term to use this: https://www.relishapp.com/vcr/vcr/docs/test-frameworks/usage-with-rspec-metadata

calumhalcrow:
@tmhammer sorry for taking so long to get back to you on this. Please read my feedback and make the relevant changes. They are summarized as follows:

- [ ] @naics_mapper -> attr_accessor :naics_mapper
- [ ] one method, and passing the year as an argument
- [ ] No cache in NaicsMapper
- [ ] rspec metadata for cassettes

calumhalcrow:
I put this on staging, but haven't re-run any imports or anything like that. If you'd like me to help do anything of that nature just let me know.

tmhammer:
@calumhalcrow  I'm not 100% sure what you're talking about with the Cassettes, do you just mean using this syntax where possible?  `describe "VCR example group metadata", :vcr do`.  I tried doing this in the V1 Trade Lead specs but was getting unexpected unhandled request errors.  I can take a closer look.

I think everything else has been fixed.

calumhalcrow:
@tmhammer I was just thinking that you might be able to do less manual configuring of cassettes by using that feature. If you ran into trouble with it then let's just leave it as is.

PR #274, opened 2015-12-23
Add aggregations to trade event lead apis response

viclim:


calumhalcrow:
Generally this looks good. I think the format of the aggregations field in the results is good.

It appears that the aggregations content isn't being populated correctly however. I'm seeing this locally:

screen shot 2015-12-24 at 3 15 15 pm

viclim:
@calumhalcrow Seems like aggregations only search within the query scope. Will have to convert trade lead and trade event query to use filtered query instead of top level filter.

spec/api/v2/trade_leads/consolidated_spec.rb
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
calumhalcrow: Is this intentional?

calumhalcrow:
@viclim the code changes required are to aggregate on the "ita_industries" field (pending @ErikSArnold's confirmation). Other than than this is looking good :+1:

spec/api/v2/trade_leads/consolidated_spec.rb
viclim: Not really. Magic comment added by emacs since the spec contains utf8 characters. Will remove it.

PR #276, opened 2016-01-07
Add flag to disable bitly in development

viclim:


calumhalcrow:
Oh how I've missed conveniently importing the CSL locally!

calumhalcrow:
@viclim can you rebase onto master, then check if tests are passing? I'm getting some failures locally, and also in CI after restarting the failed run.

calumhalcrow:
Hi @viclim, sorry for moving the goal-posts on you like this: there is another 3rd party system that we pull data from that needs to be enable-able via config. The system is "protege", and I have just merged to master the code that pulls from it. Could you rebase this on master and add that 3rd config setting?

PR #277, opened 2016-01-08
New taxonomies endpoint

tmhammer:


tmhammer:
Story:  https://www.pivotaltracker.com/story/show/110913176

config/country_mappings.yaml
```diff
@@ -52,12 +52,12 @@ United Kingdom:
...
-- !ruby/regexp /\AViet Nam\Z/i
+#Vietnam:
+#- !ruby/regexp /\AViet Nam\Z/i
 # iso_country_codes-0.6.1 lists Vietnam as one word but the next release has
 # it updated to 'Viet Nam'. I'm leaving this here to avoid having to debug it
 # again a second time.
-#Viet Nam:
-#- !ruby/regexp /\AVietnam\Z/i
+Viet Nam:
+- !ruby/regexp /\AVietnam\Z/i
```
calumhalcrow: You intend to map "Vietnam" to "Viet Nam"?

spec/api/v1/trade_leads_spec.rb
```diff
@@ -24,17 +24,25 @@
...
-    end
+  end
+
+  before do
+    allow(Date).to receive(:today).and_return(Date.parse('2015-12-18'))
+      TradeLead::Ustda.recreate_index
+      VCR.use_cassette('importers/trade_leads/ustda.yml', record: :once) do
+        TradeLead::UstdaData.new("#{Rails.root}/spec/fixtures/trade_leads/ustda/leads.xml",
+                               "#{Rails.root}/spec/fixtures/trade_leads/ustda/rss.xml").import
+      end
```
calumhalcrow: You need to do this before _each_ example?

calumhalcrow:
rubocop -a ? :stuck_out_tongue_closed_eyes:

calumhalcrow:
Other than the three items of feedback I put, looks good!

tmhammer:
Thanks @calumhalcrow!  After updating, the ISO gem seems to now only accept "Viet Nam" as the valid name.  And for the spec, putting the Date method stub in the before(:all) block caused an error, which is why I broke out the USTA portion.  Basically just need to make sure the Date stub gets run before USTDA import and I'm not sure if there's a better way.

tmhammer:
@calumhalcrow, the USTDA specs fix has already been merged.  I updated the taxonomy parser to now require the Protege URL to be passed in on the webservices side... not sure if this is the best way to handle that in development

PR #282, opened 2016-01-15
Refactoring duplicate/complex code

loren:
- Similar filter_from_fields in BusinessServiceProviderQuery and ItaOfficeLocationQuery
- Similar multi_match query with unneeded bool/must as there's only ever one child
- Created GenericScreeningList class to consolidate common behavior
- Moved similar Envirotech behavior into base_data superclass
- Moved similar TradeEvent behavior into superclass

### DIFF
--- a/app/importers/envirotech.rb
+++ b/app/importers/envirotech.rb
@@ -1,6 +1,19 @@
 module Envirotech
   include CanImportAllSources
 
+  ENVIROTECH_ISSUE_HASH = {
+    'id'                  => :source_id,
+    'name_chinese'        => :name_chinese,
+    'name_english'        => :name_english,
+    'name_french'         => :name_french,
+    'name_portuguese'     => :name_portuguese,
+    'name_spanish'        => :name_spanish,
+    'created_at'          => :source_created_at,
+    'updated_at'          => :source_updated_at,
+    'url'                 => :url,
+    'envirotech_issue_id' => :issue_ids,
+  }.freeze
+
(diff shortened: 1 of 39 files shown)


loren:
Calum, I refactored a bunch of classes resulting in about 280 fewer lines and a lot of Code Climate issues resolved (https://codeclimate.com/github/GovWizely/webservices/compare/code_climate_jan_2016#ratings). Tests still pass. Does this look reasonable?

calumhalcrow:
Sounds great :smiley: I will review this during my work day today.

app/importers/screening_list/generic_screening_list.rb
```diff
@@ -0,0 +1,8 @@
+module ScreeningList
+  class GenericScreeningList
```
calumhalcrow: Maybe just `Base`?

app/importers/trade_event/trade_event_data.rb
```diff
@@ -0,0 +1,9 @@
+module TradeEvent
+  class TradeEventData
```
calumhalcrow: `BaseData`?

I often wanted to rename the importers from *Data to *Importer. Didn't ever get around to it though.

calumhalcrow:
Looks good, thanks for doing this :smiley: 

Let me know what you think of my naming suggestions. Otherwise, I'm happy to merge this.

loren:
Thank you for looking it over! I made both name changes.

PR #290, opened 2016-01-27
Trade and World Regions for Events and Leads

tmhammer:


tmhammer:
Story:  https://www.pivotaltracker.com/story/show/111984101

calumhalcrow:
I didn't get a chance to look at this properly today, but I'd suggest digging into those ten new CC issues.

tmhammer:
I still need to do some refactoring in the Query classes and squash the commits, which I will try to get to in the next few days.  Otherwise looking good.

tmhammer:
I think this is good to go @calumhalcrow, I'm not sure why CC is complaining about the "aggregate_terms_by" since it's just a method call.  I do think the query code in general could use some refactoring, but that could be for another time.

app/importers/concerns/taxonomy_methods.rb
```diff
@@ -0,0 +1,35 @@
+module TaxonomyMethods
+  attr_accessor :taxonomy_parser
+
+  def get_missing_country(mappable_field)
+    country_array = normalize_industry('Missing Country: ' + mappable_field)
```
calumhalcrow: This line suggests that in order to lookup a "missing" country in IM, you have to pass the string "Missing Country: [COUNTRY NAME]" to the _industry_ lookup endpoint of IM. Presumably there must be industry/topic entries in IM with similar names supporting this.

If my understanding is incorrect then can you describe how this works? Otherwise, don't you think this approach is a bit  convoluted? Wouldn't it be better to map countries via another model and endpoint in IM?

app/importers/concerns/taxonomy_methods.rb
```diff
@@ -0,0 +1,35 @@
...
+    country_array.nil? ? nil : lookup_country(country_array.first)
+  end
+
+  def set_taxonomy_parser
+    self.taxonomy_parser = TaxonomyParser.new(Rails.configuration.frozen_protege_source)
+    taxonomy_parser.concepts = YAML.load_file(Rails.configuration.frozen_taxonomy_concepts)
+  end
```
calumhalcrow: Rather than dusting the importers with `set_taxonomy_parser` calls, better to lazily instantiate. Something like:

```
def taxonomy_parser
  unless self.taxonomy_parser
    self.taxonomy_parser = TaxonomyParser.new(Rails.configuration.frozen_protege_source)
    taxonomy_parser.concepts = YAML.load_file(Rails.configuration.frozen_taxonomy_concepts)
  end
  self.taxonomy_parser
end
```

app/importers/trade_lead/fbopen_importer.rb
```diff
@@ -0,0 +1,13 @@
...
+    def process_geo_fields(entry)
+      entry[:country] = IsoCountryCodes.find(entry[:country]).alpha2 rescue nil
+      entry.merge! add_geo_fields([entry[:country]])
+    end
+
+    def extract_end_date(entry)
+      [entry[:arch_date], entry[:resp_date]].reject(&:nil?).max
+    end
+  end
+end
```
calumhalcrow: I think you better give this module a different name. You've given it the same name as the module that contains the classes that you include it into (see trade_lead/fbopen_importer/full_data.rb below). I'm not sure why your code is actually working with that in place, but the phrase "circular dependencies" is already entering my head :smile:

lib/query.rb
```diff
@@ -198,6 +198,22 @@ def generate_aggregations(json)
...
 
+  def set_geo_instance_variables(options)
+    @countries = options[:countries].upcase.split(',').map(&:strip) if options[:countries].present?
+    @trade_regions = split_to_array(options, :trade_regions)
+    @world_regions = split_to_array(options, :world_regions)
+  end
+
+  def split_to_array(options, key)
+    options[key].split(',').map(&:strip) if options[key].present?
+  end
```
calumhalcrow: I don't understand why you pass the entire hash as the first arg and the key you wish to split as the second. Why not `def split_to_array(string)`?

lib/query.rb
```diff
@@ -198,6 +198,22 @@ def generate_aggregations(json)
     end
   end
 
+  def set_geo_instance_variables(options)
+    @countries = options[:countries].upcase.split(',').map(&:strip) if options[:countries].present?
```
calumhalcrow: Could `split_to_array` do this?

app/importers/concerns/taxonomy_methods.rb
tmhammer: The idea with this is to map a country onto an entry by sending in a mappable_field, such as a title or description or venue name, that will help someone decide which country should be assigned to the entry.  I prepended the field's value with "Missing Country:  " so that in Mapper, you can keyword search "Missing Country" to see the fields that need a country, and still be able to filter by source (model_name).  It's a different and more specialized use case than the Industry/Topic lookups, but the same functionality.  We could rename the industry lookup methods in IMC and Webservices to something more general.

app/importers/concerns/taxonomy_methods.rb
tmhammer: Good call on the lazy instantiation, that is much, much cleaner.  

For your second point, I think either way makes sense.  When I did the gem originally, I didn't plan for the concepts to be set at initialization time, rather they're set by the parser after calling the parse method (which is actually where these concepts come from... see lib/tasks/taxonomy.rake).  This was more of a workaround for performance reasons.

app/importers/trade_lead/fbopen_importer.rb
tmhammer: Ha, yeah not sure either why what worked, maybe there's a clear enough delineation between namespacing and mixins to where Ruby doesn't get confused by this.

PR #291, opened 2016-02-03
[#109757486] Admins can specify data source as URL

loren:
- All text entries are run through sanitize_entry method now to match behavior of existing APIs
- Removed unneeded strip_tags transformation as this will happen by default now
- Refactored sanitize_entry into its own HTMLEntityUtils module for easier reuse
- Pulled data/yaml chunks out of data_source_spec and into fixture files
- Added endpointme:import rake task for URL-based data sources with published flag set to true
- Only freshen the data when it's checksum has changed; delete any records that are no longer in the URL feed
- Generate consistent ID's for model records even if they don't have any unique_fields set


jamesisaacs:
The `url_109757486` branch was accidentally merged into master. I reset the commit and merged `url_109757486` into `staging`.

PR #300, opened 2016-02-17
Tim/salesforce

tmhammer:


tmhammer:
Upgrading the simplecov gem seems to have caused a drop in coverage unrelated to the new endpoint changes, so I've fixed these in the second commit by removing some unused code and re-enabling this Exim spec.

Also I'm not sure it would be worth the trouble to try to fix the issues CC is raising.

calumhalcrow:
@tmhammer I rearranged the SalesforceArticle importer classes so that they follow a similar arrangement to other groups of "consolidated" importers. I had to make some changes to VersionableResource in order to support this, but that's fine; they were genuine limitations of the module.

I'd like to deploy this to staging now, but to do that I need to make changes to the cookbook, specifically I need to production access creds for the SFDC API. Can you send those to me over a secure channel?

calumhalcrow:
@tmhammer let me know when you have updated the SFDC API access creds on staging so that I can finish testing this and merge.

calumhalcrow:
Ok, I'm happy to merge this at any time.

PR #302, opened 2016-02-22
Map topics

tmhammer:


jamesisaacs:
Please singularize `ita_industries` to `ita_industry` for consistency.

tmhammer:
@jamesisaacs, there is an `ita_industries` field on the Trade lead endpoints, plural to indicate there can be multiple terms.  I think if anything the singular names with arrays should change to plural but I think it's ok as is.

jamesisaacs:
@tmhammer ok. makes sense.

PR #303, opened 2016-02-23
[#114327051] Prune unused endpoints and specs

loren:
Removes 4 endpoints and some unused specs


loren:
I intentionally left this PR broken into 6 commits to reflect the 4 API endpoint removals, v1 removal, and unused tests removal. That way it will be easier to resurrect any one piece of it should that come to pass.

PR #319, opened 2016-03-17
Update rubocop config.yml

jamesisaacs:


loren:
should we run `rubocop -a` as part of this commit or just leave it for the next person to incorporate as part of their branch?

jamesisaacs:
Now seems like a good time to run `rubocop -a`.

jamesisaacs:
I just found out it `rubocop -a` introduces build errors so let's move on without running it for now.

PR #320, opened 2016-03-17
remove rubocop gem

jamesisaacs:


calumhalcrow:
But why? I always run my changes through rubocop before pushing, and ask
others to do the same when reviewing PRs.

jamesisaacs:
It's a temporary move. A breaking change from rubocop revealed some problems. I thought you should have a look. I sent you details on slack pointing to a PT story.

calumhalcrow:
Yeah, I'll take a look. I'm interested in what's causing the problem...

PR #322, opened 2016-03-18
Tim/taxonomy endpoint

tmhammer:


tmhammer:
@davidpranata  Would you mind reviewing this?  I ended up adding in some code to strip punctuation from the expanded queries in the results, just a quick fix for now.  I don't think it should effect their search behavior and it keeps them looking clean.

PR #329, opened 2016-03-30
[#110202888] Endpointme admin can consolidate multiple data sources

loren:


calumhalcrow:
@loren @jamesisaacs I see the "do not merge" label here, but I'd quite like to review this since it will allow me to take another look at the Endpoint Me code. Hope that's ok.

loren:
Have at it! I still have some work to do on data_source.rb as it's currently a D.

calumhalcrow:
As long as the user follows the instructions in [#110202888](https://www.pivotaltracker.com/projects/1085558/stories/110202888) then this works as intended. I found that if I created a consolidated endpoint first without setting its Dictionary, upon revisiting the endpoint's show page I got an error. I was unable to edit the endpoint or delete it even. But I guess that's not really a problem given the general use case of Endpoint Me (internal tool).

loren:
Thanks for taking a close look at it! I created a Tracker story to capture this issue.

PR #330, opened 2016-04-01
Query parsing for Events and Leads, related terms lookup

tmhammer:


tmhammer:
Stories:  https://www.pivotaltracker.com/story/show/115901371
https://www.pivotaltracker.com/story/show/116247993

tmhammer:
@davidpranata This should be ready for final review, you can add the "code review done" label after.  I will go ahead and merge this branch to staging at some point today.

tmhammer:
Also FYI @davidpranata, I will remove `taxonomy.rake` and related files on my next round of changes for this:  https://www.pivotaltracker.com/story/show/116468589

davidpranata:
@tmhammer 
- When searching for `Southeast asia`, it searches for the following terms:
  ```json
    ...
    { "world_regions": [ "Southeast Asia",
                         "Asia",
                         "East Asia" ]
    }
    ...
  ```
  It should be limited to `Southeast Asia` region.

- Decouple query expansion generator from the Api::V2::ItaTaxonomyController.
  The Api::V2::ItaTaxonomyController#query_expansion should not have #build_json, #build_queries, #strip_punctuation, etc.
  They should be encapsulated in something like a QueryExpansionGenerator
  ```ruby
  # Api::V2::ItaTaxonomyController#query_expansion
  ...
  @results = QueryExpansionGenerator.generate q
  ...
  ```

- Decouple QueryParser from Query classes.
  QueryParser should not be generating json queries.
  It should be a standalone module with .parse method that returns non geo name query and taxonomies.

tmhammer:
Thanks @davidpranata !  Appreciate the feedback as always.  Nice catch on that first point.  

I'm not sure what you mean about the QueryParser.  Do you just mean that the query parsing code and the parsed query generation code shouldn't be in the same place?  Separate modules?

davidpranata:
@tmhammer The [`QueryParser`](https://github.com/GovWizely/webservices/blob/dda6d084d4bcdf5343c85711df5734777fcd3582/lib/query_parser.rb) should be decoupled from [`V2::TradeEvent::Query`](https://github.com/GovWizely/webservices/blob/dda6d084d4bcdf5343c85711df5734777fcd3582/app/queries/v2/trade_event/query.rb) and [`V2::TradeLead::Query`] (https://github.com/GovWizely/webservices/blob/dda6d084d4bcdf5343c85711df5734777fcd3582/app/queries/v2/trade_lead/query.rb)

The current implementation uses `QueryParser` as a mixin in `V2::TradeEvent::Query` and `V2::TradeLead::Query` that parse the query, modify instance variables, and generate JSON.
Instead, you want `QueryParser` module that is only responsible for generating non geo names query and related terms from the original query.
The other methods in `QueryParser` such as #update_instance_variables, #generate_parsed_query, #generate_parsed_bool_child and #generate_parsed_children be in `V2::TradeEvent::Query` and `V2::TradeLead::Query`.

Your QueryParser may look something like
```ruby
#lib/query_parser.rb
module QueryParser
  def self.parse(query, types)
    ...
  end
end
```

Once `QueryParser` module becomes a standalone module, then you can write additional specs to test its behaviors.

```ruby
#spec/lib/query_parser_spec.rb
describe FbopenParser do
  describe '.parse' do
    context 'when query contains tokens that match multiple world regions' do
      let(:query) { 'Southeast Asia' }
      ...
    end
  end
end
```

tmhammer:
Ok, thanks @davidpranata that makes sense.  I would leave  #update_instance_variables, #generate_parsed_query, #generate_parsed_bool_child and #generate_parsed_children in their own module included in Lead and Event query classes to avoid the code duplication but otherwise that sounds good.

tmhammer:
@davidpranata, check out the latest commit.  There were some unrelated changes that resulted from running rubocop.

davidpranata:
@tmhammer

- You can pass `q` as an argument to the other methods in [QueryExpansionGenerator](https://github.com/GovWizely/webservices/blob/tim/taxonomy-refactor/lib/query_expansion_generator.rb). Do you need the `@@q` to remain after executing `QueryExpansionGenerator.parse` ?

- Is there plan to remove dependency on test_data.zip files in the specs where we cannot see the changes in the file without unzip and diff?

- Is test environment the only place where you set [`config.enable_related_term_lookups`](https://github.com/GovWizely/webservices/blame/ea43c0430b1202c610df4735e071a151c047ebc3/config/application.rb#L66) to false?

tmhammer:
@davidpranata 
1)  I don't think so, I just did this to make things a bit cleaner and easier to follow since q is getting referenced in many different places.
2)  test_data.zip is just what's in the test_data folder, I wasn't planning on changing this.
3)  Yes

davidpranata:
@tmhammer Please replace `@@q` class variable and pass it as an argument to the other methods in `QueryExpansionGenerator`.
Let's drop the `config.enable_related_term_lookups`. If for some reason you need `ItaTaxonomy.search_related_terms` to return an empty array, you can add `before(:all) { ItaTaxonomy.recreate_index }` in your specs.
We can deal with `test_data.zip` later.

tmhammer:
@davidpranata 
I'm not clear on why the config is not a good idea.  We don't want to rely on actual data in the ItaTaxonomy index when running specs.  That would mean we may potentially have to change every results fixture that uses the related terms lookup if we change the ItaTaxonomy test data.  It seems easier to turn it off by default, then turn it on wherever it's needed.

davidpranata:
@tmhammer 

Let's look at the [trade_event/ustda_data_spec.rb](https://github.com/GovWizely/webservices/blob/tim/taxonomy-refactor/spec/importers/trade_event/ustda_data_spec.rb).
It expects `an importer which indexes the correct documents`
It does not mention anything about ItaTaxonomy or about ItaTaxonomy.search_related_terms returning an empty array.

The [expected documents](https://github.com/GovWizely/webservices/blob/tim/taxonomy-refactor/spec/importers/trade_event/ustda_data_spec.rb) contain country names the `trade_regions` and `world_regions` are empty.
The correct behavior is to have `trade_regions` and `world_regions` populated when a valid `country name` is present.
The spec indicates that I should expect the `trade_regions` and `world_regions` to be empty at all time.

This incorrect expected result issue is widespread across this PR.
I am not sure why we want specs with the wrong expected results.
Specs with clarity and correctness are preferred over `cleaner and more efficient` specs.

tmhammer:
Ok @davidpranata,  I don't think it's necessary to have the related terms lookup functionality integrated across all specs, but fair point.  So you would recommend populating the test ItaTaxonomy index with the entire Protege source for the entire spec run, except for the specs specifically dealing with the ItaTaxonomy endpoint (which use the test_data.zip)?

davidpranata:
@tmhammer You can load the whole Protege data but I do not recommend that.

The one place where you must have .owl file is when you are writing specs for the TaxonomyParser.
The TaxonomyParser is a gem and how it works is outside the scope of webservices.
Specs for Trade event importers, trade lead importers, QueryParser depend on the documents in ItaTaxonomy but not on how the documents were added.
You can generate YAML fixtures for ItaTaxonomy load them directly into the index without having to run TaxonomyParser.

tmhammer:
Alright cool, thanks @davidpranata that sounds good.  I think what I'll do is create a YAML fixture of Countries, World Regions, and Trade Regions to populate the ItaTaxonomy test index, and that will be the default state of the index when running specs, then keep a separate smaller set of test data for the ItaTaxonomy importer and API specs.

tmhammer:
@davidpranata, latest changes are ready.  I'm all ears if you have any suggestions for improving the Taxonomy index management in rspec (see the changes in `spec_helper.rb`).  I created a story for updating the Parser like we talked about yesterday, but left it alone for now:  https://www.pivotaltracker.com/story/show/117640903.  For now I'd like to focus on getting the new FAQs out the door, which will require this being merged first.

tmhammer:
Actually @davidpranata  for the Taxonomy index in rspec I think I'm going to change it so that this part is declared specifically before each example that needs it, instead of checking every time:      
```
      ItaTaxonomy.recreate_index
      ItaTaxonomyData.new(Rails.configuration.full_protege_source,
                          YAML.load_file(Rails.configuration.full_taxonomy_geo_terms),).import
```

davidpranata:
@tmhammer You can do something like this:

```ruby
RSpec.shared_context 'ItaTaxonomy data' do
  before(:all) do
    ItaTaxonomy.create_index
    # load and index yaml
    # yaml = ...
    # ItaTaxonomy.index yaml
  end

  # always clean up after
  after(:all) { ItaTaxonomy.delete_index }
end

RSpec.describe 'some class that require ItaTaxonomy data' do
  include_context 'ItaTaxonomy data'
  # your spec
end
```

tmhammer:
Thanks @davidpranata, I have made this change.  The `full_related_terms` size is small enough to where there's not a significant performance difference in recreating and reimporting every time, vs. only doing it when needed.  If this file grows larger in the future we could always switch back to the second approach.

davidpranata:
@tmhammer I do not understand why we need
[`config.full_related_terms = "#{Rails.root}/spec/fixtures/ita_taxonomies/full_related_terms.yaml"`](https://github.com/GovWizely/webservices/blob/550408c4b8462a75b8f69d1304ba1296467ab54e/config/application.rb#L57)
when the only place that reference this config is the [`spec/support/taxonomy_data_context.rb`](https://github.com/GovWizely/webservices/blob/550408c4b8462a75b8f69d1304ba1296467ab54e/spec/support/taxonomy_data_context.rb#L5)?

How about something like this?
```ruby
# https://github.com/GovWizely/webservices/blob/550408c4b8462a75b8f69d1304ba1296467ab54e/spec/support/taxonomy_data_context.rb#L5
ItaTaxonomy.index(YAML.load_file(Rails.root.join('spec/fixtures/ita_taxonomies/full_related_terms.yaml')))
```

Am I missing something?

tmhammer:
No, you're correct, I'll update that.  Anything else?

davidpranata:
@tmhammer Please fix that one config and deploy it to staging.

PR #334, opened 2016-04-18
Exclude old AU and STATE Leads

tmhammer:


tmhammer:
Related stories:
https://www.pivotaltracker.com/story/show/117728069
https://www.pivotaltracker.com/story/show/117560015

PR #337, opened 2016-04-22
[#108383156] Add ITA Industries to Trade Events

tmhammer:


tmhammer:
Story:  https://www.pivotaltracker.com/story/show/108383156

app/importers/trade_event/ita_data.rb
```diff
@@ -6,6 +6,7 @@ class ItaData < BaseData
     include VersionableResource
 
     ENDPOINT = "http://emenuapps.ita.doc.gov/ePublic/GetEventXML?StartDT=#{Date.tomorrow.strftime('%m/%d/%Y')}&EndDT=01/01/2020"
+    CONTAINS_MAPPER_LOOKUPS = true
```
calumhalcrow: Regarding the place where you refer to the value of this constant. Take note:

```
$ irb
2.2.4 :001 > class Foo; BAR = false; end
 => false
2.2.4 :002 > Foo.const_defined?(:BAR)
 => true
2.2.4 :003 > Foo::BAR
 => false
```

PR #361, opened 2016-07-21
[#126600139] switchover trade_leads API to endpointme

loren:
- continue to support versionless route
- delete trade_leads code/tests and any other references to them
- delete other unused code (e.g., Naics mapper)
- remove restriction of 'trade_leads' as valid endpointme name


loren:
The failing code climate check seems a little unfair, as I didn't touch either V2::TradeEvent::Query or SalesforceArticle::Query. And SalesforceArticle::Query improved from a B => A, probably due to similar trade leads code getting yanked. Either way, I think trade_events is next to get migrated to endpointme, so these code climate issues will go away then.

PR #392, opened 2017-01-25
Hourly rake task to upload static files of search data to S3

tmhammer:


tmhammer:
@jamesisaacs FYI a bunch on unrelated rubocop changes have also made their way into this PR.

PR #399, opened 2017-03-13
[#141364051] CSL searcher can use countries filter when searching for an address

davidpranata:


jamesisaacs:
This fix has been deployed to production

PR #407, opened 2017-05-23
Sort param

tmhammer:


tmhammer:
@davidpranata  I'm thinking it makes more sense to make the default sort be `name:asc` since `name` is on all the endpoints, and the layered date sorting makes things unnecessarily confusing.  Any thoughts?

tmhammer:
@davidpranata I changed the mapping back.  I also bumped the Ruby version because 2.2 is no longer compatible with the gem updates of the previous PR.

tmhammer:
@davidpranata  Updated

PR #433, opened 2018-09-27
[EDSP-526] Deprecate Market Research Library endpoint

loren:


jamesisaacs:
@loren this will be merged to master and deployed on Monday 10/1, the MRL's retirement day.

